Issue #19, opened 2020-04-18
Memmory related Crash


npapan69:
Dear All,

Thanks for the great library, I'm facing problems when trying to use pandas to import my csv which has 130 rows and 110 columns. It consumes all my RAM (which is 64GB, MacBook Pro 16") and crashes. Any ideas? Up till now, I was able to use it only with csv files with a limited number of columns .

Nikos

pycaret:
Hi,

It is quite surprising that with 64GB memory a csv load fails. Can you share the size of csv file? is it dataset publicly available? if yes, I would like to try that on several pc's as well. Can you share?

Thanks.

pycaret:
One more question - are you using Python 32 bit or 64 bit?

npapan69:
I'm using MacOS Catalina and 64bit Python. The size of the csv is about 5MB with 4300 columns and 99 rows. I'm uploading here the xlsx version of it.
[test_pycaret.xlsx](https://github.com/pycaret/pycaret/files/4496923/test_pycaret.xlsx)

Thanks

npapan69:
The problem happens using JupiterNotebooks both locally and in Google Collabs

pycaret:
Hi, I will check this out and get back.

mfahadakbar:
@moezali1 can you share the details with me too? I think I experienced this issue as well, will talk to you about it

DaGuT:
I confirm the same issue on Ubuntu 19.10 with python notebook.

It wants to eat more than 400gigs of RAM for a 6mb csv file :\

Log:

> ---------------------------------------------------------------------------
> MemoryError                               Traceback (most recent call last)
>  in 
> ----> 1 exp_clf101 = setup(data = data, target = 'Label')
> 
[...]

PR #26, opened 2020-04-19
fix memory issue for features with big numbers (ID column detection)

DaGuT:
Fix for https://github.com/pycaret/pycaret/issues/19

It was caused by ID column detector.

It was creating big ranges for features that had very big numbers and then comparing such ranges with sorted features. Sometimes it results in "out of memory" problem.

Now there are no ranges. ID column is detected by increment checking. If all increments in sorted array are 1, then it's ID column.

Potential way to save more memory and power:
Use loop and compare increment by increment unless there is a difference other than 1.
It will save more memory as there will be no additional array created + when it's not ID column, loop will stop at first stages without having to calculate further increments. While now all the increments are calculated.

But I decided to keep the same "beautiful" style based on numpy and pandas functions.

___

Tested on ID column with int numbers 880:1021, column gets data type "ID Column" as it should.

### DIFF
--- a/preprocess.py
+++ b/preprocess.py
@@ -120,15 +120,13 @@ class DataTypes_Auto_infer(BaseEstimator,TransformerMixin):
         if i not in self.numerical_features:
           if sum(data[i].isna()) == 0: 
             if len(data[i].unique()) == len_samples:
-              min_number = min(data[i])
-              max_number = max(data[i])
-              arr = np.arange(min_number,max_number+1,1)
-              try:
-                all_match = sum(data[i].sort_values() == arr)
-                if all_match == len_samples:
-                  self.id_columns.append(i) 
-              except:
-                None 
+              # we extract column and sort it
+              features = data[i].sort_values()
+              # no we subtract i+1-th value from i-th (calculating increments)
+              increments = features.diff()[1:]
+              # if all increments are 0 (with float tolerance), then the column is ID column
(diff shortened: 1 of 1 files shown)


preprocess.py
```diff
@@ -120,15 +120,13 @@ def fit(self,dataset,y=None): # learning data types of all the columns
...
-                all_match = sum(data[i].sort_values() == arr)
-                if all_match == len_samples:
-                  self.id_columns.append(i) 
-              except:
-                None 
+              # we extract column and sort it
+              features = data[i].sort_values()
+              # no we subtract i+1-th value from i-th (calculating increments)
+              increments = features.diff()[1:]
+              # if all increments are 0 (with float tolerance), then the column is ID column
```
DaGuT, line 127: typo. All increments are 1! Not zero :)

pycaret: Thanks. Looking at it.

mfahadakbar:
thanks @DaGuT 
@moezali1 accept the pull request

pycaret:
@DaGuT Hi,

Can you please submit a pull request on dev-1.0.1 branch? I will take these changes in May 2020 release?

Thanks.

DaGuT:
> @DaGuT Hi,
> 
> Can you please submit a pull request on dev-1.0.1 branch? I will take these changes in May 2020 release?
> 
> Thanks.

https://github.com/pycaret/pycaret/pull/71

Issue #18, opened 2020-04-18
Dataframe constructor not properly called


BismeetSingh:
While running a simple linear regression model on kaggle using pycaret, I get the error 
DataFrame constructor not properly called on running setup. The dataset contains only two columns.

```
df=pd.read_csv("/kaggle/input/salary-data-simple-linear-regression/Salary_Data.csv")`
setup_data1 = setup(data = df, target = 'Salary', session_id=123) 
```
Logs:
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-12-c3ad091d442b> in <module>
----> 1 setup_data1 = setup(data = df, target = 'Salary', session_id=123)

/opt/conda/lib/python3.6/site-packages/pycaret/regression.py in setup(data, target, train_size, sampling, sample_estimator, categorical_features, categorical_imputation, ordinal_features, high_cardinality_features, high_cardinality_method, numeric_features, numeric_imputation, date_features, ignore_features, normalize, normalize_method, transformation, transformation_method, handle_unknown_categorical, unknown_categorical_method, pca, pca_method, pca_components, ignore_low_variance, combine_rare_levels, rare_level_threshold, bin_numeric_features, remove_outliers, outliers_threshold, remove_multicollinearity, multicollinearity_threshold, create_clusters, cluster_iter, polynomial_features, polynomial_degree, trigonometry_features, polynomial_threshold, group_features, group_names, feature_selection, feature_selection_threshold, feature_interaction, feature_ratio, interaction_threshold, transform_target, transform_target_method, session_id, silent, profile)
    955                                           target_transformation = transform_target, #new
    956                                           target_transformation_method = transform_target_method_pass, #new
--> 957                                           random_state = seed)
    958 
    959     progress.value += 1

[...]

PR #34, opened 2020-04-22
fix for dataframe constructor with 1 feature

DaGuT:
Fix for https://github.com/pycaret/pycaret/issues/18

Tested on my old big dataframe from the other my pull request + tested on the data from issue above

The problem happens at the stage when highly correlated features are removed.

Correlation between 1 feature is 1, which is not the matrix, so pandas dataframe can not be created and there is no need for further correlation analysis. 
I added extra checking to see if there is only 1 feature and return original frame if there is.

Also, to save some time for very big dataframes (as drop had to be done 3 times +1 with this fix), I added 1 variable and replaced with it those 3 places. Sorry for having 2 fixes in the same place. I know it's a bad practice, but...

### DIFF
--- a/preprocess.py
+++ b/preprocess.py
@@ -1956,9 +1956,16 @@ class Remove_100(BaseEstimator,TransformerMixin):
 
   def fit_transform(self,dataset,y=None):
     data = dataset.copy()
-    corr = pd.DataFrame(np.corrcoef(data.drop(self.target,axis=1).T))
-    corr.columns = data.drop(self.target,axis=1).columns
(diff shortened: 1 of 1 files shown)


pycaret:
@DaGuT Hi - same for this one. Can you please submit a PR on dev-1.0.1.

Thanks.

DaGuT:
> @DaGuT Hi - same for this one. Can you please submit a PR on dev-1.0.1.
> 
> Thanks.

https://github.com/pycaret/pycaret/pull/70

PR #52, opened 2020-04-26
Fixed unseen data logic for ANO101

virajpai:
If we drop the index from data in the first place, we are not actually having unseen data. Unseen data would just be 0.05 fraction of tail of original data which might be already present in data.


pycaret:
@virajpai Thank you for working on this. Would you please be able to submit a fresh PR for correction in the tutorial. Make sure you sync the `master`.

Closing this PR.

PR #71, opened 2020-04-28
fix memory issue for features with big numbers (ID column detection)

DaGuT:
Fix for https://github.com/pycaret/pycaret/issues/19

It was caused by ID column detector.

It was creating big ranges for features that had very big numbers and then comparing such ranges with sorted features. Sometimes it results in "out of memory" problem.

Now there are no ranges. ID column is detected by increment checking. If all increments in sorted array are 1, then it's ID column.

Potential way to save more memory and power:
Use loop and compare increment by increment unless there is a difference other than 1.
It will save more memory as there will be no additional array created + when it's not ID column, loop will stop at first stages without having to calculate further increments. While now all the increments are calculated.

But I decided to keep the same "beautiful" style based on numpy and pandas functions.

___

Tested on ID column with int numbers 880:1021, column gets data type "ID Column" as it should.

### DIFF
--- a/preprocess.py
+++ b/preprocess.py
@@ -120,15 +120,13 @@ class DataTypes_Auto_infer(BaseEstimator,TransformerMixin):
         if i not in self.numerical_features:
           if sum(data[i].isna()) == 0: 
             if len(data[i].unique()) == len_samples:
-              min_number = min(data[i])
-              max_number = max(data[i])
-              arr = np.arange(min_number,max_number+1,1)
-              try:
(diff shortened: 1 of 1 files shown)


pycaret:
Thanks. Merged into dev-1.0.01.

DaGuT:
This one is not merged, it was just closed. However this bug is critical
@pycaret

pycaret:
@DaGuT Thanks. Merged now.

Would you like to please give me your full name and email address to be added to contributors list on website? You can send me at moez@pycaret.org.

Thanks.

Issue #58, opened 2020-04-26
Set n_jobs


reza1615:
How can i set n_jobs=-1

pycaret:
n_jobs are by default set to -1 in all processes.

reza1615:
in the below image shows n_jobs is None
![image](https://user-images.githubusercontent.com/2202483/80313615-f0fc7d80-87b9-11ea-886b-3a62515f080f.png)

pycaret:
Actuall you are right. This is a bug. 

Thanks for pointing out.

pycaret:
Re-opened and marked as bug.

Issue #53, opened 2020-04-26
Feature request: training and predicting time to the compare_models


reza1615:
Please add training and predicting time to the compare_models table.
![image](https://user-images.githubusercontent.com/2202483/80307397-d95fcd80-8796-11ea-99f5-7c2cac7cfe6b.png)

pycaret:
Thank you for your contribution. I will mark this as an enhancement for future release.

reza1615:
@pycaret You removed the history https://github.com/pycaret/pycaret/graphs/contributors
I don't contribute here

pycaret:
@reza1615 Is it only showing contribution on master branch?

reza1615:
I had [this](https://github.com/pycaret/pycaret/pull/74) commit  after you reuploaded the file all history removed [here ](https://github.com/pycaret/pycaret/commits/dev-1.0.1/pycaret/classification.py)

PR #74, opened 2020-04-28
n_jobs =-1 + training_time + MCC

reza1615:
* added n_jobs =-1 bug #58 
* added training time + MCC #53 
* Highlight Mean Row as Green


pycaret:
Thanks. Merged into dev-1.0.1.

pycaret:
@reza1615 It would be good to add log loss metric as well.

pycaret:
@reza1615 we should also add the green color on mean rows and n_jobs fix in pycaret.regression.

PR #77, opened 2020-04-28
update pycaret version

DaGuT:
meh, just having 1 more free commit with version update :\

Since pycaret will become version 1.0.1 with this branch, I updated setup.py to match its new version


pycaret:
Thanks.

Issue #69, opened 2020-04-28
pip installation error


kevin7lou:
Would you pls help to check `pip` installation error? Thanks!
```
abc@jackloumbp  ~  pip install pycaret                                           ✔  12:43:02

Collecting pycaret
  Downloading pycaret-1.0.0-py3-none-any.whl (188 kB)
     |████████████████████████████████| 188 kB 640 kB/s
Collecting cufflinks==0.17.0
  Downloading cufflinks-0.17.0.tar.gz (81 kB)
     |████████████████████████████████| 81 kB 3.1 MB/s
Collecting catboost==0.20.2
  Downloading catboost-0.20.2-cp37-none-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl (10.6 MB)
     |████████████████████████████████| 10.6 MB 902 kB/s
Collecting plotly==4.4.1
  Downloading plotly-4.4.1-py2.py3-none-any.whl (7.3 MB)
     |████████████████████████████████| 7.3 MB 7.0 MB/s
Requirement already satisfied: IPython in ./anaconda3/lib/python3.7/site-packages (from pycaret) (7.13.0)
Collecting yellowbrick==1.0.1
  Downloading yellowbrick-1.0.1-py3-none-any.whl (378 kB)
     |████████████████████████████████| 378 kB 5.5 MB/s
Collecting pyod
  Downloading pyod-0.7.8.2.tar.gz (92 kB)
     |████████████████████████████████| 92 kB 588 kB/s
Collecting scikit-learn==0.22
  Downloading scikit_learn-0.22-cp37-cp37m-macosx_10_6_intel.whl (10.9 MB)
     |████████████████████████████████| 10.9 MB 3.8 MB/s
Requirement already satisfied: numpy in ./anaconda3/lib/python3.7/site-packages (from pycaret) (1.18.1)
Collecting lightgbm==2.3.1
  Downloading lightgbm-2.3.1-py2.py3-none-macosx_10_9_x86_64.macosx_10_10_x86_64.macosx_10_11_x86_64.macosx_10_12_x86_64.macosx_10_13_x86_64.macosx_10_14_x86_64.macosx_10_15_x86_64.whl (679 kB)
     |████████████████████████████████| 679 kB 4.0 MB/s
Collecting spacy
[...]

Issue #48, opened 2020-04-25
interpret_model() problem with XGBoost regression


nasospat:
I was trying interpret_model with XGBoost for a regression model and had this error
 
![shap_xgboost](https://user-images.githubusercontent.com/26347750/80281748-061fd080-8716-11ea-9980-28ee6bb91187.PNG)

I searched for it and seems to be a known issue of shap-0.32.1. 
https://github.com/slundberg/shap/issues/887
I updated to shap-0.35.0 and the problem was fixed. Maybe you should consider not using the shap-0.32.1 version.

pycaret:
Thanks for identifying. Will do it in next release.

Issue #15, opened 2020-04-16
Problem installing pycaret


klahrich:
Hi everyone,

I had some issues pip-installing pycaret. Some of my other libraries use the same underlying packages as pycaret, but require more recent versions.

Here are the error messages I get:

ERROR: prettierplot 0.1.1 has requirement scikit-learn>=0.22.1, but you'll have scikit-learn 0.22 which is incompatible.
ERROR: mlmachine 0.1.3 has requirement catboost>=0.22, but you'll have catboost 0.20.2 which is incompatible.
ERROR: mlmachine 0.1.3 has requirement scikit-learn>=0.22.1, but you'll have scikit-learn 0.22 which is incompatible.
ERROR: mlmachine 0.1.3 has requirement shap>=0.35.0, but you'll have shap 0.32.1 which is incompatible.
ERROR: mlmachine 0.1.3 has requirement xgboost>=1.0.2, but you'll have xgboost 0.90 which is incompatible.

Would it be a possibility to update your requirements file ?

Thank you

LinxiFan:
Agreed, the `requirements.txt` seems to be very restrictive. Is there a reason to require those packages to be the exact version number?

KaitoKikuhara:
I thought to solve that problem setup.py have to be changed with the requirements file. because of requirements in setup.py too.

PR #78, opened 2020-04-28
fix for strict requirements

DaGuT:
Accidentally updated minimal version for xgboost in https://github.com/pycaret/pycaret/pull/76 . This one is replacement for that :)
__
Requirements are too strict. Nothing important was deprecated in newer versions of packages. Old versions do not work on all computers (like xgboost can't be easily compiled and installed on latest versions of MAC OS). So I've just replaced == with >= for people to install more recent versions. It will also help solve some compatibility issues with other packages that require more recent versions

Fix for
https://github.com/pycaret/pycaret/issues/15
https://github.com/pycaret/pycaret/issues/69
https://github.com/pycaret/pycaret/issues/48

Tested on MAC and only basic features. If possible, more tests are required. However, nothing should be broken since no features (used in this package) are deprecated, so it can be approved and merged.

Sadly, it will interfere with https://github.com/pycaret/pycaret/pull/75 (no longer true, since he let this update cover the problem)


pycaret:
Thanks. Merged in dev-1.0.1 branch.

Ideally we should point out to requirements.txt file instead of reproducing in install_requires in setup.py. What do you think? 

Is requirements.txt also updated based on the changes made in setup.py?

DaGuT:
> Merged in dev-1.0.1 branch.
> 
> Ideally we should point out to requirements.txt file instead of reproducing in install_requires in setup.py. What do you think?

Hm... That might be a good idea. Honestly, I've never been paying attention. But what I can see 
in e.g. pandas and numpy they dont have requirements.py, since everything is written in setup.py

> 
> Is requirements.txt also updated based on the changes made in setup.py?
As I mentioned before, requirements.txt usually is not present (but I've checked only two big packages :D). Anyway, it's not update automatically.

Nevertheless, here is autofetching from requirements.py in setup.py https://github.com/pycaret/pycaret/pull/81

PR #81, opened 2020-04-28
Moved setup requirements to requirements.txt only

DaGuT:
Before we had to change requirements in two files, now it's done only in requirements.txt, so that people who do not want to install it as package can just simply load it from git and install with requirements.txt. That's the reason we dont move everything to just setup.py

Though, you might keep it as it is now. I've never been paying attention how it's done, so I just did as it's said in https://github.com/pycaret/pycaret/pull/78

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,1 @@
-include README.md LICENSE
+include README.md LICENSE requirements.txt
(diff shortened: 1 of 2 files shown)


pycaret:
merged.

PR #83, opened 2020-04-29
+ category_encoders

reza1615:
Adding category_encoders


reza1615:
@moezali1: you remove the history! I don't contribute here
https://github.com/pycaret/pycaret/graphs/contributors

PR #85, opened 2020-04-29
Adding Training time +mean in green + n_job=-1

reza1615:
* added n_jobs =-1 bug #58 
* added training time  #53 
* Highlight Mean Row as Green

### DIFF
--- a/regression.py
+++ b/regression.py
@@ -2,8 +2,6 @@
 # Author: Moez Ali <moez.ali@queensu.ca>
 # License: MIT
 
-
-
 def setup(data, 
           target, 
           train_size=0.7,
@@ -1114,7 +1112,7 @@ def setup(data,
     
     #sample estimator
     if sample_estimator is None:
-        model = LinearRegression()
+        model = LinearRegression(n_jobs=-1)
     else:
         model = sample_estimator
         
@@ -1507,9 +1505,6 @@ def setup(data,
         
         return X, y, X_train, X_test, y_train, y_test, seed, prep_pipe, target_inverse_transformer, experiment__
 
-
-
-
 def create_model(estimator = None, 
                  ensemble = False, 
                  method = None, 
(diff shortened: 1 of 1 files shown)


reza1615:
Attached a notebook unit test for the regression.py functions
[Regresion unit test.zip](https://github.com/pycaret/pycaret/files/4551841/Regresion.unit.test.zip)

pycaret:
@reza1615 Thanks.

pycaret:
@reza1615 Thanks for unit test as well. both for clf and reg.

I just realized that I don't think so there is value of adding training time in all functions other than compare_models because there is no value of seeing the training time by fold in create_model, tune_model, ensemble_model. Only compare_models should have training time. 

Also to avoid the training time being confused with metrics, do you think it would be a good idea to round it to 2 decimals?

Let me know what do you think?

reza1615:
> @reza1615 Thanks for unit test as well. both for clf and reg.
> 
> I just realized that I don't think so there is value of adding training time in all functions other than compare_models because there is no value of seeing the training time by fold in create_model, tune_model, ensemble_model. Only compare_models should have training time.
> 
> Also to avoid the training time being confused with metrics, do you think it would be a good idea to round it to 2 decimals?
> 
> Let me know what do you think?

I am totally agree.
Also i suggest to  set others decimals to 4

reza1615:
In my opinion Training time for other function is useful user can realize how long does it take to train a model. for example after 3 folder he/she decide to cancel learning based on elapsed time and changing some parameters to speed up the whole process.

pycaret:
@reza1615 for that purpose the monitor that runs on on top will always show time remaining. so even after 1 fold person can choose to decide to cancel the training. 

For compare_models training time on monitor top is not good because it doesn't tell us by model. Having them as a column makes sense there.

I think if we force round = 4, does that mean MAE for integer would also show 4 decimals? for example 24000 would become 24000.0000?

pycaret:
@reza1615 Just realized all the metrics are already rounded to 4 decimals. The reason it shows more than 4 decimals in your regression unit test is probably because you have used gold dataset which actually goes beyond 4 decimals. Does it make sense to you?

PR #86, opened 2020-04-29
MAPE function+ decimal format

reza1615:
* Rewrite calculate_mape  based on https://github.com/scikit-learn/scikit-learn/pull/15007/files/a64758ef54fa1dfd8ea741dd693f1c166078f010#diff-4873478c97f8e898b176ad552a4d96a2
It should be multiply with 100
* formating training time and MAPE

### DIFF
--- a/regression.py
+++ b/regression.py
@@ -2,6 +2,13 @@
 # Author: Moez Ali <moez.ali@queensu.ca>
 # License: MIT
 
+def calculate_mape(actual, prediction):
+    import numpy as np
+    epsilon = np.finfo(np.float64).eps
+    mape = np.abs((prediction - actual) / np.maximum(np.abs(actual), epsilon))
+    output_errors = np.average(mape, axis=0) * 100
+    return np.average(output_errors)
+
 def setup(data, 
           target, 
           train_size=0.7,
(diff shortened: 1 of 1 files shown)


pycaret:
@reza1615 I think it is implied for MAPE that it is a percentage, just like R2. At the moment 11.2% MAPE is shown as 0.112. I believe your change will make it 11.2. is that right?

reza1615:
Yes my change shows 11.2%

pycaret:
@reza1615 Hi. I think we don't need to show % sign. It will become inconsistent because AUC, Accuracy, Recall, Precision, R2 all these metrics are presented without % sign. Can you take back the change of format change of MAPE. I like the idea of how you have centralized calculate_mape function.

pycaret:
@reza1615 decimal format for time is also fine.

reza1615:
@pycaret comparing to AUC, Accuracy, Recall, Precision, R2,
MAPE is the only metric which has percentage in its name and definition

pycaret:
@reza1615 Hey Reza, what I actually meant by removing % is to remove the multiplication to 100. The formula should be same as what it was.

reza1615:
@pycaret done

reza1615:
@pycaret did you check the code?

reza1615:
@pycaret : you removed the history! https://github.com/pycaret/pycaret/graphs/contributors
I don't contribute here

pycaret:
@reza1615 This graph is only for master branch only.

reza1615:
@pycaret I had [this](https://github.com/pycaret/pycaret/pull/74) commit  after you reuploaded the file all history removed [here ](https://github.com/pycaret/pycaret/commits/dev-1.0.1/pycaret/classification.py)

PR #95, opened 2020-05-01
fix typos clustering beginner tutorial 

stexaci:
I fixed a few typos and revised some sentences.

Also, I replaced the code below:

`data = dataset.sample(frac=0.95, random_state=786).reset_index(drop=True)`
`data_unseen = dataset.drop(data.index).reset_index(drop=True)`

with:

`data = dataset.sample(frac=0.95, random_state=786)`
`data_unseen = dataset.drop(data.index).reset_index(drop=True)`
`data.reset_index(drop=True, inplace=True)`

because if we use the former code, there will be an overlap between data and data_unseen.


stexaci:
@pycaret 
Thank you for the encouragement! I didn't notice any changes in metrics in this particular tutorial, but if the same code is used in other tutorials as well (regression, classification, etc.), then, I suspect there would definitely be cases where the values of the metrics in the text should be replaced with the new ones. If you have more important things to do, I would try to fix all of them within a week.

PR #97, opened 2020-05-02
Anomaly detection tutorial beginner fix typos

stexaci:
I fixed a few typos in the text and edited some sentences. Also, I modified the code for splitting the dataset into `data` and `data_unseen`, so that there would not be any overlap between those two, i.e.
I replaced:
`data = dataset.sample(frac=0.95, random_state=786).reset_index(drop=True)`
`data_unseen = dataset.drop(data.index).reset_index(drop=True)`
with:
`data = dataset.sample(frac=0.95, random_state=786)`
`data_unseen = dataset.drop(data.index).reset_index(drop=True)`
`data.reset_index(drop=True, inplace=True)`

 In this particular tutorial, the changed results on the new `data_unseen` did not affect any values in the text, so there is are no cross-referencing issues.


pycaret:
Thanks for working on it.

Merged. Have you ran all the notebooks on your local environment after making the changes?

pycaret:
@stexaci 

Hi, the split issue you identified is in almost all the tutorials. As such it has to be fixed in other tutorials as well.

Thanks for helping.

stexaci:
Yes, I ran the notebooks, but when the results in the estimation outputs differ from the ones in the text (for instance, "AUC"), I change the corresponding values in the text as well.

PR #99, opened 2020-05-03
Use Cases

Riazone:
I have added Use Cases Folder where people can add their Use Cases. 
I have added my topic of Gold Prediction. It has currently two set of experiments, Regression and Classification. Both have their separate set of folders and a common Data folder


Riazone:
Adding Use Cases

pycaret:
Thanks. Merged.

Issue #135, opened 2020-05-14
Could not convert STR column to FLOAT when this column has missing values


FaridAhamadaGunners:
I load a dataset (8000,22) with a string column contains float numbers. So the setup() function infer that this column is categorical because this column contains whitespace characters. 
So when i force the setup() function with the parameter numeric_features set to with this column 

I encountered this error : ValueError: could not convert string to float 

When i debug the code i see that in the script preprocess.py you use the function .astype(float64) for conversion. I think you should use the pandas function to_numeric() with the parameter errors set to 'coerce' and the parameter downcast to 'integer'  or 'float or whatever. This function with theses settings will convert any columns in numeric and will force the unrecognized values to NaN.

pycaret:
@FaridAhamadaGunners Can you submit the PR to dev-1.0.1 branch?

FaridAhamadaGunners:
Yes, i will do it

PR #154, opened 2020-05-18
Fix the float conversion bug #Issue#153

FaridAhamadaGunners:
Basic code : 
- If the input column contains string character or 'commas' separators the conversion doesn't work

New code  :
- In order to facilitate the user experience. Automating this task as well as possible is beneficial
- The new one can convert an object column to float column by converting the numbers to float and set to NaN the unrecognized characters.

### DIFF
--- a/pycaret/preprocess.py
+++ b/pycaret/preprocess.py
@@ -201,12 +201,19 @@ class DataTypes_Auto_infer(BaseEstimator,TransformerMixin):
         except:
           data[i]=dataset[i].apply(str)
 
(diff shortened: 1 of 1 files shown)


FaridAhamadaGunners:
correcting tag Issue : #135

pycaret:
## Reviewer required for this change.

pycaret:
@FaridAhamadaGunners Can you please submit this PR on `2.2-refactor`. 

@Yard1 is leading the refactoring sprint and would be happy to change there if he thinks it is necessary and not yet addressed.

Thank you for your help and time.

PR #217, opened 2020-06-12
Update nlp.py

CodeFighter03:
Here the autosentiment library return the positive, negative or neutral text pieplot,number of text type list, percentage and the array of type text list.


pycaret:
@CodeFighter03 Thanks for submitting this. I will start looking on it in next 2 weeks.

PR #279, opened 2020-07-10
Raghibshams456 patch 1

Raghibshams456:
replace string with list of string of parameter blacklist in module regression.py


pycaret:
Change already taken in PyCaret 2.0.

Thanks

PR #292, opened 2020-07-19
Preventing CatBoost models from creating files

GLevV:
Should prevent CatBoostClassifier and CatBoostRegressor from writing files to disk. The downside is that catboost methods and functions for visualisations and snapshots will be unavailable.
"If set to “False”, the snapshot and data visualization tools are unavailable." from docs.

### DIFF
--- a/classification.py
+++ b/classification.py
@@ -4240,7 +4240,7 @@ def tune_model(estimator = None,
                       #'ctr_border_count':[50,5,10,20,100,200]
                       }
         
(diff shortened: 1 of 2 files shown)


pycaret:
@GLevV Thanks for submitting. I don't think so it's anymore necessary, since we have added **kwargs in **create_model** base function. All other functions will adopt to it in future. So if somebody wants to prevent writing catboost files to local, it can be done by directly passing argument in **create_model**

Feel free to work on other open issues. 

Thanks

PR #303, opened 2020-07-23
Don't split boolean category columns

Yard1:
First of all, I would like to thank everyone for this awesome library.

One issue I noticed was that categorical columns with just two possible values (boolean categories) were being split into two columns during dummyfication. For example, `GENDER` with values `[F, M]` would be split into `GENDER_F` and `GENDER_M`. This seems completely unnecessary, and may have a negative impact on the model. I have tweaked the `Dummify` class in `preprocessing` to account for those.

### DIFF
--- a/pycaret/preprocess.py
+++ b/pycaret/preprocess.py
@@ -1128,9 +1128,28 @@ class Dummify(BaseEstimator,TransformerMixin):
     self.target = target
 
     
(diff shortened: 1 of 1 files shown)


Yard1:
This seems to cause some issues with certain datasets - I will investigate further.

Issue #335, opened 2020-08-03
Apply Python code formatter


daikikatsuragawa:
How about using a code formatter like [black](https://github.com/psf/black) to unify the format?

pycaret:
@daikikatsuragawa Thanks. I will have a look at this. Thanks for submitting PR.

pycaret:
@Yard1 Should we close this issue?

Yard1:
Yeah

PR #336, opened 2020-08-03
Apply black

daikikatsuragawa:
Apply Python code formatter ([black](https://github.com/psf/black)) #335

As a first step, the settings are:

```toml:pyproject.toml
[tool.black]
line-length = 120
```


pycaret:
@daikikatsuragawa Thanks for submitting this. I will hold on this one for now and will request you to run this again right before next release.

pycaret:
@Yard1 @daikikatsuragawa Would it also make sense to create a GitHub action for this? Can we schedule this once in a day? or once in a week? What are your thoughts?

Yard1:
@pycaret IMO we should format everything after the refactor (which will cause a lot of conflicts anyway) and start enforcing styling after that with a github action preventing people from merging PRs unless they follow the style.

daikikatsuragawa:
Automatic formatting is also possible.
I think It supposed to be pull request time and push time.
Some of the projects I've been involved with have worked fine and are well developed.

There is also a mechanism called pre-commit.

Yard1:
It is possible, but if we use an action to do it, then a bot will need to make a commit after every merge/push, which will quickly make the history unreadable, while pre-commit hooks need to be set-up locally by the developer themselves, which is not something we can enforce.

Issue #262, opened 2020-07-02
request to add categorical labels for multi_class classification but not in numerical categories.


sathishkumar369:
Hi pycaret team!

when I gave prediction on unseen data, the output come up with label values.
I want it in categorical(original form) form but not in label form.

hope you will understand my points by looking into images.

Actual inputs mentioned in the figure and it's a multiclass one

![Screenshot (244)](https://user-images.githubusercontent.com/40235632/86370484-70686a80-bc9d-11ea-9535-8373fadd0c85.png)

predicted one mentioned in the below figure

![Screenshot (242)](https://user-images.githubusercontent.com/40235632/86370528-7d855980-bc9d-11ea-9444-4744d2952364.png)

pycaret:
Thanks for the suggestion @sathishkumar369.

Tagging this as an enhancement.

ingebert:
Hi Team,
thanks for making pycaret so useful and awesome, and thanks Satish for the suggestion.
I would highly appreciate this feature, too!
If I understand this right, this would be an implementation of sklearn.preprocessing.LabelEncoder.inverse_transform(transformed_label_as_in_prediciton)
It looks like the "sklearn label encoder" is already used in pycaret code but could be made accessible by some option in setup? I´m not good enough in coding (yet :) ) and hope this can be done easily or with workaround by making the encoded/decoded array accessible (see picture).
![image](https://user-images.githubusercontent.com/68894423/89284087-7fe92380-d64e-11ea-89a5-581eb27d3f85.png)
I tried already to access it with get_config('prep_pipe') but couldn´t find it.

Yard1:
I added a PR for this, https://github.com/pycaret/pycaret/pull/348

pycaret:
This issue is resolved. 

Thanks for working on this @Yard1

PR #348, opened 2020-08-04
Categorical labels for classification predict_models()

Yard1:
Adds an option to classificaiton `predict_models()` to output labels as original categories, instead of outputting them in an encoded integer form. Default behavior is unchanged.

References https://github.com/pycaret/pycaret/issues/262

### DIFF
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -10630,16 +10630,6 @@ def predict_model(estimator,
     """
     exception checking ends here
     """
-
-    # function to replace encoded labels with their original values
-    # will not run if categorical_labels is false
-    def replace_lables_in_column(label_column):
-        if not categorical_labels:
-            return
-        _, dtypes = next(step for step in prep_pipe.steps if step[0] == "dtypes")
-        if dtypes and hasattr(dtypes, "replacement"):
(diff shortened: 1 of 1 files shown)


Yard1:
Whoops, good catch. Looks like I missed one when renaming.

ingebert:
Another issue popped up when trying to run the model in production, running ` load_model()`  and ` predict_model()` only -  without executing ` setup()`  and  `create_model()` before  :
In that case `prep_pipe ` is not defined. 
To make it work, I suggest accessing the `estimator` list from the `predict_model` function argument.
As there are several operations in between (i.e.  estimator.pop(), where the pipeline steps are dropped) , 
a copy has to be made in the beginning.
I suggest following code changes (or something similar):
image

Yard1:
Will take a look, thanks

Yard1:
@ingebert Pushed an update, please check if it helps.

ingebert:
Works fine now, thank you for the quick update

pycaret:
@Yard1 Can you sync `master` and submit a fresh PR to incorporate the change in `predict_model` in `classification.py`.

Thanks for working on this.

Issue #353, opened 2020-08-04
Reducing risk for data leakage


sparvaneh:
It sees that pre-processing (e.g. imputation) is done before splitting data to train and test. Current workflow increase risk of data leakage as mean/median of the whole dataset will be used for imputation while in real application only mean/median of training set should be used for imputation. 
Easy solution could be doing data split first and then pre-process/prepare data.

As another suggestion, separation of creating train/test data from setup under a separate module in the future versions could make it easier to use externally provided training and test sets.

pycaret:
@sparvaneh Thanks for suggestion. Adding **enhancement** tag for now.

Yard1:
I created a PR for this, https://github.com/pycaret/pycaret/pull/360. In addition, some code has been refactored for easier maintenance (identical code blocks in if elses have been moved outside).

pycaret:
@Yard1 Can you submit a separate PR just for the split change?

Yard1:
I'll redo this PR @pycaret

pycaret:
@Yard1 As discussed, we will hold-on for this one for `2.1` release and give some thought to fix it during refactoring.

pycaret:
## Issue parked for `pycaret==2.2`. Assigned to @Yard1 

### Please do not work on this issue.

pycaret:
@sparvaneh 

This issue has been fixed in the `pycaret-nightly` release. Please install pycaret-nightly using `pip install pycaret-nightly`. Feel free to open a new issue, in case the problem persists. This fix will be published to PyPi as `pycaret==2.2.0` on October 28, 2020.

PR #360, opened 2020-08-05
Preprocess after splitting

Yard1:
In accordance with https://github.com/pycaret/pycaret/issues/353, preprocessing for classification and regression now occurs after train-test splitting. The pipeline is ran separately for both sets. 

Additionally, the code has been refactored:
- Generating preprocessing pipeline doesn't fit and transform data right away (added a parameter for that, `fit_transform_and_return`, to `Preprocess_Path_One`, default behavior is unchanged)

I have ran some additional tests, and everything works as expected.

Given that this is a temporary fix, it is not recommended to merge at this time.


Yard1:
My bad. I pushed a new commit. Could you check if it works now, @sparvaneh ? I am fitting data twice, before and after splitting, so that the target is transformed too. Otherwise, it would only be transformed for the train set.

clgarciga:
@Yard1 @pycaret What is the status of this? Wondering because preprocessing before splitting is currently causing an error with #503...

Yard1:
We held off on this for now as it would require some deep changes to make it work properly. I think we want to do it for 2.2.

clgarciga:
@Yard1 Okay, thanks!

Issue #337, opened 2020-08-03
List of unit tests that need to be added


daikikatsuragawa:
- [x] test_arules.py  #338
- [x] test_datasets.py #365
- [x] test_nlp.py https://github.com/pycaret/pycaret/commit/773f14329a7ec8208aded68afae272c7808acfec
- [x] test_preprocess.py #409
- [x] test_utils.py #393

pycaret:
@daikikatsuragawa Thanks for working on this. 

Will be waiting for rest of the tests.

daikikatsuragawa:
I will proceed step by step. After completing the process, we will review each test and improve the quality.

daikikatsuragawa:
@pycaret #365 

I propose two solutions to the `test_nlp.py` problem.

- Add `python -m spacy download en` to `.travis.yml`.
- Add `python -m spacy download en` to `.requirements.txt`.

daikikatsuragawa:
For the time being, all Python files have been tested.

pycaret:
@daikikatsuragawa Thank you so much.

PR #365, opened 2020-08-05
Create test_datasets.py

daikikatsuragawa:
#337

### DIFF
--- a/pycaret/tests/test_datasets.py
+++ b/pycaret/tests/test_datasets.py
@@ -1,0 +1,17 @@
+import os, sys
+sys.path.insert(0, os.path.abspath(".."))
+
+import pytest
+import pycaret.datasets
+
+def test():
+    # loading list of datasets
+    data = pycaret.datasets.get_data('index')
+
+    # loading dataset
+    credit = pycaret.datasets.get_data('credit')
+
+    assert 1 == 1
+    
+if __name__ == "__main__":
+    test()


pycaret:
@daikikatsuragawa Thanks. When you get a chance can you have a look at test case for nlp.py. The problem there is that build fails on travis because spacy english model must downloaded using command line for test to run. I don't know how to do that on runtime in travis.

Anyways, apart from nlp.py you can also build test cases for preprocess.py. It would be very extensive though as we might have to create multiple tests as there are many functionalities for preprocessing pipeline. So coverage will be important.  

Another easy one would be test case for **arules.py**

daikikatsuragawa:
@pycaret  Thanks for the merge! I have seen similar issues with `nlp.py`.
Is adding `python3 -m spacy download en` to `.travis.yml` a solution to this problem?

daikikatsuragawa:
If there is no problem, add the following code and edit `.travis.yml`.

```py
import os, sys
sys.path.insert(0, os.path.abspath(".."))

import pytest
import pycaret.nlp
import pycaret.datasets

def test():
    # loading dataset
    data = pycaret.datasets.get_data('kiva')

    # init setup
    exp_nlp101 = pycaret.nlp.setup(data = data, target = 'en', session_id = 123)
    
    # create model
    lda = pycaret.nlp.create_model('lda')

    # assign model
    lda_results = pycaret.nlp.assign_model(lda)

    # evaluate model
    pycaret.nlp.evaluate_model(lda)

    assert 1 == 1
    
if __name__ == "__main__":
    test()
```

daikikatsuragawa:
@pycaret 
I verified. This way you will succeed!

pycaret:
@daikikatsuragawa I am trying now. Will add `test_nlp.py` shortly.

Issue #358, opened 2020-08-05
Refractor compare_models() to use create_model()? 


Yard1:
It seems to me that it would be a good idea to move out all model-generating code out of `compare_models()`, and instead use `create_models()` internally. Right now, the code that essentially does the same thing is duplicated in two places, making maintenance and extensibility harder. If `compare_models()` was to be refactored like that, it would be much easier to add new models, and also, allow for variations of the function that eg. compare tuned or existing models.

I'd be happy to try my hand at this if you feel it is a good idea. It would be a big change, and it may introduce some new issues, but I think it is necessary for the project going forward.

pycaret:
@Yard1 

It all started with this one single PR. :) 

Closing this now...

PR #371, opened 2020-08-05
Refactor compare_models()

Yard1:
Refactors `compare_models()` in classification and regression modules in order to:
* Use `create_models()` internally, with **unchanged** display (required modification of create_models(), however default functionality is unchanged)
* Accept completed models to be compared with `estimator_list` parameter - it is possible to now run code such as:
```python
top3 = compare_models(n_select=3)
top3_tuned = [tune_model(x) for x in top3)
compare_models(estimator_list=top3+top3_tuned) # will create a score grid of passed models, allowing for easy comparison of untuned and tuned ones. Also works with ensembled and blended models, though stacked models are currently unsupported
```
image (the index is on purpose, as it is the same as the index of elements in the passed list)
* Cut down on the code, general cleanup of some parts

This will allow maintenance to be much easier, as `compare_models()` will not need to be edited aside from adding some names to dicts (which should be global, to be honest). As a side effect, `compare_models()` is now faster, as it doesn't need to recreate models again for saving best. Later, I would like to add an option to compare models with train set instead of CV.

Also made it so that `create_models()` shows the TT in score grid (so that it can be shown in `compare_models()`. Could probably be extended to other functions.

References https://github.com/pycaret/pycaret/issues/358

### DIFF
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -1975,6 +1975,7 @@ def create_model(estimator = None,
                  cross_validation = True, #added in pycaret==2.0.0
                  verbose = True,
                  system = True, #added in pycaret==2.0.0
+                 compare_monitor = None,
                  **kwargs): #added in pycaret==2.0.0
 
     """  
@@ -2049,6 +2050,9 @@ def create_model(estimator = None,
 
     system: Boolean, default = True
     Must remain True all times. Only to be changed by internal functions.
+
+    compare_monitor: DataFrame, default = None
+    Must remain None all times. Only to be changed by compare_models()
 
     **kwargs: 
     Additional keyword arguments to pass to the estimator.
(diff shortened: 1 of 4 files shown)


Yard1:
@pycaret Could you take a look? I think that this is a very good and helpful change. I'm more of a software engineer than data scientist, so such kind of changes are up my alley.

pycaret:
@Yard1 Thanks for doing that. I am in 100% agreement that the code requires some deep cleaning. I am not a software engineer so I am guilty of writing code that is not the most optimal. However, there are couple of things I want to do before taking on these major changes. In following order:

- Adopt to Stacking Classifier instead of in-house function in stack_models
- Remove create_stacknet in PyCaret 2.1 (already committed)
- Change behavior of save_model, instead of returning list. Model object should be appended in sklearn.pipeline so that we can use it for pmml conversion. The reason it is like that is because of stack_models and create_stacknet function.
- Some other minor bug fix
- Convert repository from private to organization
- Add colloborators / members into the project (like yourself) 
- Develop processes and add reviewers in place to implement governance process.

I expect to do all this in next couple of weeks. Once I have a good start on these items we can either colloborate on how to take on the changes you have submitted through PR.

In the mean time, can you submit a separate PR for the part where you have refactored code in setup function to avoid multiple if and else. I can merge that PR sooner.

Yard1:
That works for me. The fact changing the way stacked models work is planned is why I didn't bother to add support for them. I will open up a new PR with some general code cleanup that doesn't change any functionality.

Yard1:
I'm closing this for now as the functionality in this PR will be introduced during the planned general refactor.

PR #385, opened 2020-08-07
Added the cardiotocography dataset.

dubeyakshat07:
Link to the original dataset and its details: https://archive.ics.uci.edu/ml/datasets/Cardiotocography

2126 fetal cardiotocograms (CTGs) were automatically processed and the respective diagnostic features measured. The CTGs were also classified by three expert obstetricians and a consensus classification label assigned to each of them. Classification was both with respect to a morphologic pattern (A, B, C. …) and to a fetal state (N, S, P). Therefore the dataset can be used either for 10-class or 3-class experiments.


pycaret:
@dubeyakshat07 also please update index.csv in /datasets with the meta deta of dataset you just uploaded.

dubeyakshat07:
> @dubeyakshat07 also please update index.csv in /datasets with the meta deta of dataset you just uploaded.

I have updated the index.csv in /datasets with the meta data of the 'CTG.csv' dataset.

Issue #374, opened 2020-08-06
Automated documentation / readthedocs account needed


pycaret:
# Help needed

Looking for somebody who can help develop automated documentation with theme. Currently, documentation is manually managed on website through wordpress template. It's a lot of work to maintain. We should create automated documentation using sphinx and set up a workflow to generate documents automatically with PR requests. 

Some work has already been done. Check closed PR and /doc folder in the repo. The problem with current work is that the formatting is completely out. Although structure should remain similiar / or if you want you can build new structure using pycaret.org/guide as building blocks.

Thanks for your help. For discussion reach out moez.ali@queensu.ca

Yard1:
If nobody picks this up I can take a look in a day or two.

pycaret:
@Yard1 Thanks

Yard1:
What's the formatting issue exactly? Could you show me an example of how it should look like? Or should I look at the current site for guidance?

Yard1:
Ok, yeah, it seems that the issue is that the current docstrings don't follow a format recommended by Sphinx. I'll see if this is something that can be fixed with Sphinx configuration itself, but I fear that we may need to rewrite all the docstrings to one standardized format. While we're at it, we could add typing to all the methods, too.

Yard1:
Ok, found the issue. The formatting for docstrings is missing indents. For example:

bad:
```
    Parameters
    ----------
    optimize : string, default = 'R2'
    Other values you can pass in optimize param are 'MAE', 'MSE', 'RMSE',
    'RMSLE', and 'MAPE'. 

    use_holdout: bool, default = False
    When set to True, metrics are evaluated on holdout set instead of CV.
```
good:
```
    Parameters
    ----------
    optimize : string, default = 'R2'
        Other values you can pass in optimize param are 'MAE', 'MSE', 'RMSE',
        'RMSLE', and 'MAPE'. 

    use_holdout: bool, default = False
        When set to True, metrics are evaluated on holdout set instead of CV.
```
I'll add indents where necessary and open a PR. Sphinx configuration itself seems to be fine.

Yard1:
https://github.com/pycaret/pycaret/pull/390

If you want me to, I can set up an account on readthedocs as well - though I am not sure whether it will require some repo permissions or not.

pycaret:
@Yard1 One question. The sequence of modules in the documentation and the sequence of functions inside each module? Is it automatically determined by the position of code or is it something we can control? Can we have the following order of modules in documentation

[...]

PR #390, opened 2020-08-07
Fix docstrings & doc generation

Yard1:
Fixed all docstrings, aside from those in Preprocessing, to conform to numpy standard, so that sphinx can generate docs without issues. Regenerated docs. Ready to be put on readthedocs.

In the future, all documentation **must** conform to numpy standard, that is:
https://numpydoc.readthedocs.io/en/latest/format.html
https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html

(Mostly) Fixes https://github.com/pycaret/pycaret/issues/374


pycaret:
@Yard1 Thank you so much. I am going to reply to your email soon.

pycaret:
@Yard1 Can you check the documentation of `create_model` function in documentation in any of the module. Some formatting change is required their. Name of models should appear as list.

Yard1:
@pycaret Yeah, sorry, I forgot to mention that those tables need to be reworked into something else - bullet point lists, probably, since we can't use a proper table there. Would that be fine?

pycaret:
Yea. Lets change them into something some bullet list so its readable.

Yard1:
Will do. Expect a commit very soon.

PR #418, opened 2020-08-12
Feature/gcp azure np docs

amjadraza:
# PR: Model Deployment to GCP and Azure

Below features have been added.

1. In `deploy_model` and `load_model` function support of deploying model on Azure and GCP
2. These modifications have been implemented in `classification.py` and `regression.py` modules.

> Test Google colabs are included in `examples/`. Tested well.


pycaret/classification.py
```diff
@@ -9638,6 +9687,32 @@ def deploy_model(model,
     - Default Region Name (can be seen under Global settings on your AWS console)
     - Default output format (must be left blank)
 
+    For GCP users:
+    --------------
```
Yard1, line 9691: I am pretty sure you cannot use headers like that inside doc sections - I'd just remove it (`--------`)

PR #443, opened 2020-08-14
Rearrange Sphinx documentation sections

Yard1:
Rearranges the documentation to be in the following order of modules:

1. Classification
2. Regression
3. Clustering
4. Anomaly Detection
5. Natural Language Processing
6. Association Rule Mining
7. Datasets

Such things can be changed by editing the index.rst file.

https://github.com/pycaret/pycaret/issues/374#issuecomment-673874912


pycaret:
Thanks @Yard1 Would you also know how the function level is controlled under each section? Is it linked with the way code is written or you can actually define the sequence, for eg. setup, compare_models, create_model, tune_model .......

Yard1:
@pycaret Pretty sure the order is the same as in the .py file

Issue #304, opened 2020-07-24
GPU acceleration for scikit-learn via H2O4GPU 


MattiaVerticchio:
`pycaret` is great, however, there’s no GPU support. I’m using it Google Colab and it can be very slow with large datasets. Can I suggest using [H2O4GPU](https://github.com/h2oai/h2o4gpu) as a replacement for scikit-learn? It’s available as a [Python package](https://pypi.org/project/h2o4gpu/).

> H2O4GPU is a collection of GPU solvers by H2Oai with APIs in Python and R. The Python API builds upon the easy-to-use scikit-learn API and its well-tested CPU-based algorithms. It can be used as a drop-in replacement for scikit-learn (i.e. `import h2o4gpu as sklearn`) with support for GPUs on selected (and ever-growing) algorithms. H2O4GPU inherits all the existing scikit-learn algorithms and falls back to CPU algorithms when the GPU algorithm does not support an important existing scikit-learn class option.

Yard1:
h2oai seems to have issues with python > 3.6, doesn't it?

pycaret:
@MattiaVerticchio Can you have a look at code on dev-1.0.1 branch? Maybe you would like to work on including this?

MattiaVerticchio:
> h2oai seems to have issues with python > 3.6, doesn't it?

Support for Python >3.6 is been worked on [right now](https://github.com/h2oai/h2o4gpu/pull/842).





> @MattiaVerticchio Can you have a look at code on dev-1.0.1 branch? Maybe you would like to work on including this?

Sure, I’ll  replace `sklearn` with `h2o4gpu` and test it on the tutorial examples. 👍

MattiaVerticchio:
I swapped `sklearn` with `h2o4gpu` [for supported calls](https://github.com/pycaret/pycaret/compare/dev-1.0.1...MattiaVerticchio:dev-1.0.1) and tried to `compare_models()`. There’s an overall 20% reduction in comparing times and `GradientBoostingClassifier` had a drop of ~90% in training time when running the Beginner Classification tutorial.

In my opinion, it could be a good improvement in training times, especially for large datasets, but it should be furtherly tested. I had an issue with measuring some error metrics with the `LogisticRegression` model when running `compare_results()`.

You can test it in [this Colab Notebook](https://colab.research.google.com/drive/1w6OvzPhKkEWyaJ5_WT8HgfdGnv1w0Dad?usp=sharing
).

pycaret:
[...]

PR #462, opened 2020-08-17
GPU accelerated rbfsvm, rf, gbc, xgboost

MattiaVerticchio:
I added `rbfsvm`, `rf`, `gbc`, `xgboost` GPU acceleration in `pycaret.classification.compare_models()` using the global `gpu_param`.

### DIFF
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -2049,21 +2049,36 @@ def compare_models(blacklist = None,
     
     logger.info("Importing libraries")
 
-    #import sklearn dependencies
-    from sklearn.linear_model import LogisticRegression
+    # Import dependencies
+    if gpu_param:
+        from cuml.svm import SVC
+        from h2o4gpu import RandomForestClassifier
+        from h2o4gpu import GradientBoostingClassifier
+        # Uncomment when RAPIDS 0.15 is released
+        # https://github.com/rapidsai/cuml/issues/2491
+        # from cuml.neighbors import KNeighborsClassifier
+    else:
+        from sklearn.svm import SVC
+        from sklearn.ensemble import RandomForestClassifier
+        from sklearn.ensemble import GradientBoostingClassifier
+        # Uncomment when RAPIDS 0.15 is released
+        # https://github.com/rapidsai/cuml/issues/2491
+        # from sklearn.neighbors import KNeighborsClassifier
+
+    # Delete knn import when RAPIDS 0.15 is released
+    # https://github.com/rapidsai/cuml/issues/2491
     from sklearn.neighbors import KNeighborsClassifier
+
+    from sklearn.linear_model import LogisticRegression
+    from sklearn.linear_model import SGDClassifier
     from sklearn.naive_bayes import GaussianNB
     from sklearn.tree import DecisionTreeClassifier
-    from sklearn.linear_model import SGDClassifier
-    from sklearn.svm import SVC
     from sklearn.gaussian_process import GaussianProcessClassifier
     from sklearn.neural_network import MLPClassifier
     from sklearn.linear_model import RidgeClassifier
-    from sklearn.ensemble import RandomForestClassifier
     from sklearn.discriminant_analysis import QuadraticDiscriminantAnalysis
     from sklearn.ensemble import AdaBoostClassifier
-    from sklearn.ensemble import GradientBoostingClassifier
-    from sklearn.discriminant_analysis import LinearDiscriminantAnalysis 
+    from sklearn.discriminant_analysis import LinearDiscriminantAnalysis
     from sklearn.ensemble import ExtraTreesClassifier
     from sklearn.multiclass import OneVsRestClassifier
     from xgboost import XGBClassifier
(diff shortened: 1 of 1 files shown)


pycaret:
@MattiaVerticchio Thanks for working on this. Some points to start discussion:

- For `lightgbm`, `xgboost`, `catboost` we can use the built-in GPU support in the algorithms.
- `svm` and `knn` can be imported from `cuml`. I have noticed you created placeholder for `knn`, that is super helpful, thanks for that.
- `gbc` and `rf` is imported using `h204gpu`. Is this package available on `pip` and `conda`. We are in the process of creating a conda install for pycaret. Just wanna make sure, this won't be a problem. Also in the documentation, we will have to cover that the installation for `cuml` and `h204gpu` is optional and to use `use_gpu` param in `setup`, one must install these libraries separately. The way I do it is to add them in `requirements.txt` as a commented text.
- How do you plan to handle the `gpu_param` in `create_model`, `tune_model` and `blend_models` (when all is selected).

MattiaVerticchio:
> * For `lightgbm`, `xgboost`, `catboost` we can use the built-in GPU support in the algorithms.

Exactly. GPU support for those estimators is built-in and enabled through `device='gpu'`, `tree_method='gpu_hist'` and `task_type='GPU'` parameters. I didn’t enable those flags in `CatBoostClassifier` because there were speed issues I need to investigate, as stated [here](https://github.com/pycaret/pycaret/issues/304#issuecomment-674826497).

> * `svm` and `knn` can be imported from `cuml`. I have noticed you created placeholder for `knn`, that is super helpful, thanks for that.

I also tested SVM - Linear Kernel through `cuml.MBSGDClassifier` but I got worse training times. For the same reason as `CatBoostClassifier` I didn’t include it in the pull request. Since this implementation uses mini-batches to accelerate the algorithm, we need to properly tune the `batch_size` parameter to efficiently parallelize the workload and result in an improvement over `sklearn`.

About `knn`, there’s [a bug](https://github.com/rapidsai/cuml/issues/2491) that will be solved in cuML 0.15 release that prevents scoring on large datasets, hence I opted for a placeholder to be removed when it’ solved.

> * `gbc` and `rf` is imported using `h204gpu`. Is this package available on `pip` and `conda`. We are in the process of creating a conda install for pycaret. Just wanna make sure, this won't be a problem. Also in the documentation, we will have to cover that the installation for `cuml` and `h204gpu` is optional and to use `use_gpu` param in `setup`, one must install these libraries separately. The way I do it is to add them in `requirements.txt` as a commented text.

Yes, `h2o4gpu` is available [through Conda](https://github.com/h2oai/h2o4gpu).

H2O4GPU deals with `GradientBoostingClassifier` and `RandomForest` by redirecting object calls to a properly configured [`xgboost`](https://github.com/h2oai/h2o4gpu/blob/master/src/interface_py/h2o4gpu/solvers/xgboost.py), while cuML leverages its own [CUDA libraries](https://github.com/rapidsai/cuml/blob/9d5eabbea5ff58b3520ea351b9b65e3eaf6fc4a6/python/cuml/ensemble/randomforestclassifier.pyx). Also, cuML has its own [`RandomForestClassifier`](https://docs.rapids.ai/api/cuml/stable/api.html#random-forest) but it
> is an early release of the cuML Random Forest code. It contains a few known limitations:
>
> GPU-based inference is only supported if the model was trained with 32-bit (float32) datatypes. CPU-based inference may be used in this case as a slower fallback.
>
> Very deep / very wide models may exhaust available GPU memory. Future versions of cuML will provide an alternative algorithm to reduce memory consumption.

and I would wait until it’s stable before introducing it in place of H2O4GPU’s one. I prefer cuML implementation over H2O4GPU since the first one is actively developed by Nvidia with low level CUDA kernels. Unfortunately, there is no `cuml.ensemble.GradientBoostingClassifier` implementation yet, the H2O4GPU is the only available.

> * How do you plan to handle the `gpu_param` in `create_model`, `tune_model` and `blend_models` (when all is selected).

There could be conditional imports and object creation, but it’s not very elegant and maintainable to have a lot of `if/else` statements. I like [this idea](https://github.com/pycaret/pycaret/issues/304#issuecomment-674829042) for centralizing the choice of the models.

clgarciga:
@MattiaVerticchio Quick question on this: my understanding about RAPIDS packages is that only Linux is supported. Will this cause an error if someone is trying to use this on their Windows machine? And if so, is there a way to handle that exception so that it defaults to using sklearn?

pycaret:
@clgarciga This is a great question. @MattiaVerticchio Same question. Would it work on windows?

MattiaVerticchio:
RAPIDS isn’t available on Windows, a fallback to `sklearn` can be easily implemented like this:
```python
try:
    import cuml
except ImportError:
    import sklearn
```

clgarciga:
@MattiaVerticchio Could you incorporate that try-except block in your PR?

Yard1:
@MattiaVerticchio Are there any other models we could add from cuML?

MattiaVerticchio:
> @MattiaVerticchio Are there any other models we could add from cuML?

There are a few models already used in PyCaret that could be added:
- `cuml.LinearRegression`
- `cuml.LogisticRegression`
- `cuml.Ridge`
- `cuml.Lasso`
- `cuml.ElasticNet`
- `cuml.ensemble.RandomForestClassifier`
- `cuml.svm.SVC`
- `cuml.svm.SVR`
- `cuml.neighbors.KNeighborsClassifier`
- `cuml.neighbors.KNeighborsRegressor`

Also, there are:
- `cuml.MBSGDClassifier`
- `cuml.MBSGDRegressor`

which are slightly different implementations of `SGDClassifier` and `SGDRegressor`.

Besides, there are a lot of routines that could accelerate preprocessing, metrics, and utilities.
Are you working on the centralized model repository?

Yard1:
Indeed. I will contact you shortly.

Yard1:
@MattiaVerticchio I am having issues with using RandomizedSearchCV on GPU-enabled xgboost. Could you try it yourself? Pass an XGBClassifier to tune_models. It takes an extremely long time and often crashes.

MattiaVerticchio:
> Could you try it yourself?

I just tried, [here’s the notebook](https://colab.research.google.com/drive/1s98XWZNWkvaDEBPl0HWNE-rPZpzjFLt7?usp=sharing).
It seems that `tune_model` doesn’t get `tree_method = gpu_hist` from `**kwargs`.

Yard1:
@MattiaVerticchio could you do `pip install git+https://github.com/Yard1/pycaret.git@refactor` instead? That way you will get the dev version.

MattiaVerticchio:
I think the issue here is that we’re passing a cloned estimator trained with `tree_method = gpu_hist`,  to `tune_model`, which trains the model without the `gpu_hist` parameter and runs on CPU.

In the `refactor` branch we can call `tune_model(xgboost, **kwargs)`, however, by doing this I get an error from passing the `kwargs` to `RandomizedSearchCV` (that doesn’t have a `kwargs` parameter) instead of passing the parameter dictionary to the estimator.

Yard1:
Shouldn't the parameter stay the same after cloning?

Yard1:
Ok, resolved the issue. The `max_depth` parameter was set way too high, causing memory crashes.

MattiaVerticchio:
> Shouldn't the parameter stay the same after cloning?

Oh yes, you’re right, now it works. 👍 



> Ok, resolved the issue. The `max_depth` parameter was set way too high, causing memory crashes.

Nice. Also, I saw `n_estimators` explicitly set up as parameter for grid search.
I usually set it really high (E.G. `9999999`) and use early stopping on cross-validation data to find the optimal number of trees without overfitting on the training set.

Yard1:
Sadly, not every algorithm we have supports early stopping. Still, people can pass their own grids, so I don't think it will be an issue.

Issue #432, opened 2020-08-13
Introduction of GitHub Actions


daikikatsuragawa:
I think it is a good idea to run automated unit tests at a commit time. Currently, this project uses travis, but I feel that it has a higher affinity for development using GitHub. In some cases, you may use two together as a trial. Please consider.

Example: 
https://github.com/stanfordmlgroup/ngboost

pycaret:
@daikikatsuragawa Happy to do that. Assigning this to you.

pycaret:
@Yard1 can help if you need @daikikatsuragawa

Yard1:
I've got some experience with Github Actions, so let me know if you need a hand @daikikatsuragawa

daikikatsuragawa:
Thank you! !! @pycaret @Yard1

daikikatsuragawa:
#440

As a first step, I reproduced the work of Travis in GitHub Actions.

pycaret:
@daikikatsuragawa Thanks. I lack experience in GitHub actions. 

Does this mean we will discontinue travis? What is the advantage of using GitHub actions to perform automatic testing compared to travis?

pycaret:
@daikikatsuragawa One thing I would like to achieve through GitHub actions is automatically publish the changes to `pycaret-nightly` pip package. Is this something we can easily implement?

daikikatsuragawa:
@pycaret 
> Does this mean we will discontinue travis?

No, there is nothing in particular.

If the things you can do with GitHub Actions are fulfilling, I think that they may be combined and integrated.
I did the same job, hoping it would be helpful for that time.

After that, I thought that it was appropriate as the first step.

daikikatsuragawa:
Probably we can.

I think that there is something in the marketplace that can be helpful, so please check.

Are there any procedures for reference?
> automatically publish the changes to pycaret-nightly pip package.

daikikatsuragawa:
I would like to make a list of what I expect from A.
In some cases I would like to make it an independent issue.

- Auto format
- automatically publish the changes to pycaret-nightly pip package

Yard1:
Both are possible. pip package for nightly can be configured to update on every push to master or once per day.

Yard1:
[...]

PR #470, opened 2020-08-18
Nightly publish workflow

Yard1:
The following changes will put in place a workflow that:
- Will run every 24h on midnight zulu time
- Will check whether the latest commit on master that passed CI has been made in the last 24 hours
- If yes, will build and publish pycaret-nightly using that commit
- Otherwise, end action

Will require a PyPi token to be set as `pypi_password` secret. Instructions on how to do it are here - https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github

Several files needed to be edited in order to facilitate smooth building and publishing. The main change is that both setup.py files will get the version from pycaret/utils.py, in order to make sure it is in one place.

https://github.com/pycaret/pycaret/issues/432

### DIFF
--- a/.github/CI/check_latest_commit_for_nightly.py
+++ b/.github/CI/check_latest_commit_for_nightly.py
@@ -1,0 +1,77 @@
+from datetime import datetime, timedelta
+import sys
+import traceback
+import requests
+
+# Author: Antoni Baum (Yard1)
+# Helper script to see if it's worth pushing to nightly, to be used in a GitHub Workflow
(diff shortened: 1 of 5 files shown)


pycaret:
@Yard1 I have added new secret `pypi_password`.

Yard1:
@pycaret Should be good for merge, then. We'll know if it works when GMT-midnight comes :)

PR #473, opened 2020-08-18
Update python-package.yml

daikikatsuragawa:
Added the code so that the trigger works at the timing of pull request.

https://github.com/pycaret/pycaret/compare/master...daikikatsuragawa:patch?expand=1#diff-044e1b0c2cace416aaa3b7f25bcef572R3

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:


pycaret:
@daikikatsuragawa There is a conflict in this. Can you resolve it or submit a new PR?

Yard1:
Yeah, I renamed `python-package.yml` to `test.yml` as well as changed `build` to `test` inside so that it's clear what it does. You could just quickly make a new PR using up-to date master @daikikatsuragawa

Issue #394, opened 2020-08-08
Plot Model-how to Set image resolution and output high quality image


lcg29:
I want to plot high-quality images, it doesn't seem to be in the current guidance. In addition, plot style also can't find. 
Thanks.

pycaret:
@lcg29 It's not possible at the moment. Please feel free to propose what kind of change would be required inside `plot_model` function to allow high-quality images. We are using `yellowbricks` and `plotly` as backend for visualization. I think `yellowbricks` uses `matplotlib` which allows resolution to be modified.

lcg29:
Yeah，set ` yellowbricks` uses `matplotlib`  is a good idea. Thanks.  

I think `plot_model(modelname, dpi=600)`,  dpi of png maybe an excellent function.

pycaret:
@lcg29 Would you like to work on this?

lcg29:
Thanks for your invitation. I learned Python for less than a month....you know lack of ability. So I can't give you a definite answer. If I have worked out a solution, I will reply to you in time.

pycaret:
@Yard1 Do you want to have a look at this one when you get a chance? 

User story : Can we add `dpi` parameter in `plot_model` to control the high resolution of plots.

Yard1:
I cna take a look

ifyoungnet:
how to set dpi or scale when using interpret_model()?

PR #475, opened 2020-08-18
scale parameter for plot_model

Yard1:
Add `scale` parameter to plot_model to allow for rendering of plots in higher resolution (aside from NLP, which only has interactive plots).

I chose `scale` instead of `dpi` because plotly.express doesn't allow setting of dpi and only allows for pixel values. Other plots have their dpi scaled.

Fixes https://github.com/pycaret/pycaret/issues/394


pycaret:
@Yard1 Thanks. This is great. I am now testing it.

Issue #458, opened 2020-08-16
Questions about the part that calculates MAPE of utils.py


daikikatsuragawa:
In the test of utils.py(test_util.py), the output of MAPE is the data frame shown below.

```shell
[102 rows x 1 columns]
Label   NaN
medv    NaN
dtype: float64
```

I think it is desirable to get a numerical value, but how about that?

Currently, the test is responding as follows.
https://github.com/pycaret/pycaret/blob/276366a0f0d294910f26b95c5da88dc0925a9c1e/pycaret/tests/test_utils.py#L97-L99

daikikatsuragawa:
The following is for reference.

```py
mape = pycaret.utils.check_metric(actual, prediction, "MAPE")
print(actual)
print(prediction)
print(type(mape))
```

```shell
     medv
0    28.2
1    23.9
2    16.6
3    22.0
4    20.8
..    ...
97   22.9
98   44.8
99   21.7
100  10.2
101  15.4

[102 rows x 1 columns]
       Label
0    29.0434
1    24.3722
2    17.1263
3    21.3939
4    21.0816
..       ...
97   22.9265
98   45.4065
99   22.0341
100  12.2415
101  15.0220

[102 rows x 1 columns]
<class 'pandas.core.series.Series'>
```

Yard1:
What is returned for other metrics?

pycaret:
This is what I get for MAPE:

![image](https://user-images.githubusercontent.com/58118658/90346529-a9636100-dff7-11ea-86b5-0d794b50955c.png)

It's same for all other metrics in `pycaret.utils.check_metric`.

Does that help?

daikikatsuragawa:
Thank you very much. It will be helpful.

> What is returned for other metrics?
In other calculations, the numbers were output safely. The following tests have passed.
https://github.com/pycaret/pycaret/blob/master/pycaret/tests/test_utils.py#L86-L99

daikikatsuragawa:
It is the following state when I am corrected with reference to the comment…

```py

    # provisional support
    actual = actual.dropna(axis=0, how="any")
    actual = actual.reset_index()
    actual=actual.drop("index", axis=1)
    prediction = prediction.dropna(axis=0, how="any")
    prediction = prediction.reset_index()
    prediction=prediction.drop("index", axis=1)
    actual = actual.medv
    prediction = prediction.Label
    print(actual)
    print(prediction)
    # check metric(regression)
[...]

PR #476, opened 2020-08-18
Fix MAPE check_metric returning NaN, enable test

Yard1:
Fix `check_metric` for MAPE returning NaN.

Fixes https://github.com/pycaret/pycaret/issues/458

### DIFF
--- a/pycaret/tests/test_utils.py
+++ b/pycaret/tests/test_utils.py
@@ -107,9 +107,9 @@ def test():
     assert isinstance(rmsle, float)
     assert rmsle >= 0
     mape = pycaret.utils.check_metric(actual, prediction, "MAPE")
-    # provisional support
-    # assert isinstance(mape, float)
-    # assert mape >= 0
+    #provisional support
+    assert isinstance(mape, float)
+    assert mape >= 0
 
     assert 1 == 1
 
--- a/pycaret/tests/test_utils.py
+++ b/pycaret/tests/test_utils.py
@@ -107,7 +107,6 @@ def test():
     assert isinstance(rmsle, float)
     assert rmsle >= 0
     mape = pycaret.utils.check_metric(actual, prediction, "MAPE")
-    #provisional support
     assert isinstance(mape, float)
     assert mape >= 0
 
--- a/pycaret/utils.py
+++ b/pycaret/utils.py
@@ -91,11 +91,11 @@ def check_metric(actual, prediction, metric, round=4):
 
     elif metric == 'MAPE':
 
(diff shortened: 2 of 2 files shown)


Sherry40931 (commented):
Is MAPE implementation wrong? It's MPE (Mean Percentage Error), not MAPE (Mean Absolute Percentage Error)

It should be:
```python
result = (np.fabs((actual.iloc[:,0] - prediction.iloc[:,0])/actual.iloc[:,0]))[mask].mean()
```
Your commit:
```python
result = (np.fabs(actual.iloc[:,0] - prediction.iloc[:,0])/actual.iloc[:,0])[mask].mean()
```

Yard1:
The logic here has been taken from sklearn, aside from masking 0 values.

Sherry40931:
The code of mape from sklearn:
```python
epsilon = np.finfo(np.float64).eps
mape = np.abs(y_pred - y_true) / np.maximum(np.abs(y_true), epsilon)
output_errors = np.average(mape, weights=sample_weight, axis=0)
```

[source code](https://github.com/scikit-learn/scikit-learn/blob/e21319fee/sklearn/metrics/_regression.py#L197)
[document](https://scikit-learn.org/dev/modules/generated/sklearn.metrics.mean_absolute_percentage_error.html)

Yard1:
Interesting, thanks. I must have misremembered. I'll update it.

Yard1:
@Sherry40931 This method of calculating MAPE has been depreciated by a new one introduced in PyCaret 2.2, therefore there's nothing to change. The new method does indeed uses the same logic as sklearn - I just confused the two.

The code for the new method used since 2.2:
```python
        def mean_absolute_percentage_error(
            y_true, y_pred, sample_weight=None, multioutput="uniform_average"
        ):
            y_type, y_true, y_pred, multioutput = _check_reg_targets(
                y_true, y_pred, multioutput
            )
            check_consistent_length(y_true, y_pred, sample_weight)
            mask = y_true != 0
            y_true = y_true[mask]
            y_pred = y_pred[mask]
            mape = np.abs(y_pred - y_true) / np.abs(y_true)
            output_errors = np.average(mape, weights=sample_weight, axis=0)
            if isinstance(multioutput, str):
                if multioutput == "raw_values":
                    return output_errors
                elif multioutput == "uniform_average":
                    # pass None as weights to np.average: uniform mean
                    multioutput = None

            return np.average(output_errors, weights=multioutput)
```

Sherry40931:
I see. I am using 2.1.2 so I got different MAPE, I will update it.

Issue #431, opened 2020-08-13
Enhancement of test


daikikatsuragawa:
For each function, it is necessary to test it in detail.

For example, in test_utils.py:

```py
    accuracy = pycaret.utils.check_metric(actual, prediction, "Accuracy")
    assert accuracy >= 0
    assert accuracy <= 1
```

pycaret:
@daikikatsuragawa Agree. I have merged all the changes you submitted yesterday and today. Let's continue to improve unit tests.

Yard1:
We could consider preparing datasets that are sure to give consistent results with a set session ID, and use them for testing, to make sure the output of the functions is as expected.

pycaret:
@Yard1 This sounds good. I plan to discuss this with you in the meeting but I guess the idea would be to release `2.1` in the next 10 days and then start major refactoring after that for `2.2` release.

Once we close all items added into project `2.1 Authors sprint`, we will be ready for the release.

Would love to wrap up the GitHub actions thing before that.

pycaret:
@daikikatsuragawa Do you consider this complete for this sprint?

daikikatsuragawa:
At the very least, I want to complete all return type checks for the tests.
I think we can do it in the same way as before.

Also, I dare to solve this problem somehow. I can't solve it. #458

Yard1:
@daikikatsuragawa opened a PR fixing it https://github.com/pycaret/pycaret/pull/476

daikikatsuragawa:
@Yard1 Thanks!!

daikikatsuragawa:
> At the very least, I want to complete all return type checks for the tests.
> I think we can do it in the same way as before.

This seems to have been resolved.

pycaret:
@daikikatsuragawa Great. I am planning to drop the `2.1` release next week. Let me know if you need any help to finish this.

daikikatsuragawa:
No type checking is done for the libraries used to create a particular model. This is because we may use another library in the future. For example, I might change A to another library that does the same thing.

Therefore, I don't think it is mandatory for the test to fail in such cases. Because it is not the essence. Therefore, I am doing this. Based on that idea, the general type check is completed.

For the rest, I want to check the evaluation value and finish it for the time being.

However, the test is very profound...

pycaret:
[...]

PR #479, opened 2020-08-18
Update test_classification.py

daikikatsuragawa:
#431


pycaret:
@daikikatsuragawa Build failed.

daikikatsuragawa:
In future development, the model may be created using another library of the same method. Therefore, I removed some type testing.

Issue #495, opened 2020-08-20
Classification dataset with groups


clgarciga:
I've looked through the datasets included in pycaret and maybe I overlooked it but I can't find a classification dataset with non-i.i.d. data. Was wondering if there was one with groups (e.g. dataset of repeated measures on different patients - here the groups would be the patients) already included, or if anyone can think of one off the top of their head that is available online somewhere. If not I have an idea for one I can create and contribute but I figure it's probably better to pull one off the internet somewhere if it exists.

pycaret:
@clgarciga Would be good to include that in `dataset` folder and also `index.csv` must be updated with metadata.

PR #502, opened 2020-08-20
Add us_presidential_election_results data

clgarciga:
#495


clgarciga:
The objective is to predict which party wins each state based primarily on the Democratic candidate's poll average, and the Democratic candidate's poll average margin over the Republican candidate's poll average margin (i.e. Dem Poll Avg - Rep Poll Avg). The grouping variable is` cycle`.

clgarciga:
Data was created by me based on the following data sources:
- Wikipedia
- https://github.com/fivethirtyeight/data/tree/master/polls
- https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/42MVDX

PR #503, opened 2020-08-20
Predetermined and group train/test split + group CV

clgarciga:
Work in progress. To be updated shortly.

### DIFF
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -4,10 +4,10 @@
 # Release: PyCaret 2.1
 # Last modified : 19/08/2020
 
-def setup(data,  
-          target,   
-          train_size = 0.7, 
-          sampling = True, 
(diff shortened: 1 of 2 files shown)


pycaret:
@clgarciga This is a good one. Tests are failing now so I will hold on for merging but I receive a lot of requests every day to allow flexibility in `train_test_split`. This is what this PR is doing. Great job.

clgarciga:
@pycaret Thanks! I'm holding off on `TimeSeriesSplit ` until after I'm done here. Also, this PR currently includes `GroupKFold`.

clgarciga:
@pycaret Does preprocessing potentially drop rows from the data? From the docs it seems like this is what `remove_outliers` does? If so, would any other params also potentially trigger dropping rows?

PR #536, opened 2020-08-25
2.2 refactor - Add GPU enabled models

Yard1:
Adds support for [cuML](https://github.com/rapidsai/cuml) models as well as catboost and xgboost.


Yard1:
@MattiaVerticchio could you take a look, please?

pycaret/classification.py
```diff
@@ -909,8 +911,9 @@ def setup(
         raise TypeError("html parameter only accepts True or False.")
 
     # use_gpu
-    if type(use_gpu) is not bool:
-        raise TypeError("use_gpu parameter only accepts True or False.")
+    use_gpu.lower() if isinstance(use_gpu, str) else use_gpu
+    if use_gpu != "force" and type(use_gpu) is not bool:
+        raise TypeError("use_gpu parameter only accepts 'Force', True or False.")
```
MattiaVerticchio, line 927: `Force` and `force` are inconsistent.

Yard1: Yeah, I suppose it would be better to keep it consistent. Fixed.

MattiaVerticchio:
Here are some issues I got in `compare_models` and how to solve them:

- Some classes, like `LogisticRegression`, `MBSGD`, and `SVC` need `float` values to fit the models. They all have a `convert_dtype = True` parameter in their methods (`fit`, `predict`, `predict_proba`, ...) to convert features as needed.
- `SVC` supports by default the `predict_proba` method, without the need to pass `probability = True` when creating the object, like in `sklearn`.
- When setting `use_gpu = 'Force'` we are enforcing cuML ≥ 0.15, hence we can use `KNeighborsClassifier`, since the prediction bug, as of now, is fixed in 0.15.

Besides, I somehow cannot `exclude = ['svm']` in `compare_models`.

Yard1:
Thanks. 

- I think dtype conversion is already being done by default in cuML 0.15. I am testing to see if it's possible to have the target variable encoded as float64 instead of int64 - if it works fine, then we can just change preprocessing instead.
- That argument will be left in place so that both sklearn and cuML `SVC` can use the same args dict
- Thanks, I will add it.

Will check what's up with that exclude.

Yard1:
Excluding works fine for me. Maybe you seem to think it's the case due to cuML Ridge Regression being a specially configured `cuml.MBSGDClassifier`.

MattiaVerticchio:
> * I think dtype conversion is already being done by default in cuML 0.15. I am testing to see if it's possible to have the target variable encoded as float64 instead of int64 - if it works fine, then we can just change preprocessing instead.
> * That argument will be left in place so that both sklearn and cuML `SVC` can use the same args dict

I just checked after installing cuML 0.15 and it works well. 👍



> Excluding works fine for me. Maybe you seem to think it's the case due to cuML Ridge Regression being a specially configured `cuml.MBSGDClassifier`.

That’s right, it works.

I’ll try to tweak some parameters and see if there is a speedup in the new models.

PR #544, opened 2020-08-26
Inline ternary operator and boolean expressions

MattiaVerticchio:
Simplified `None` checks with an inline ternary operator and boolean expressions.


Yard1:
@MattiaVerticchio Good idea, thanks. More PRs like this are appreciated, too :+1:

PR #545, opened 2020-08-26
2.2 refactor - Rework model and metric definitions

Yard1:
Model and metric definitions have been reworked to use special classes, dynamically converted into dataframe rows. This change allows for:

- Better modularity
- Error checking when adding new models
- Standardization


Yard1:
Added cuML Random Forest Classifier.

PR #560, opened 2020-08-27
Fix the float conversion for declared list of numeric features 

FaridAhamadaGunners:
Basic code in preprocess.py:
* if the input column contains numbers separate by a 'commas' the "astype('float64')" function doesn't work because he couldn't convert string to float
* if the input column contains some string or missing values the "astype('float64')" function doesn't work because he couldn't convert string to float

New code :

In order to facilitate the user experience. Automating this task as well as possible is beneficial. The new one can : 
- Convert an object column containing numbers separate by commas. By replacing commas to point
- Convert an object column containing some string characters or missing values to float column by converting the numbers to float and set to NaN the unrecognized characters (whitespace, missing values...)

### DIFF
--- a/pycaret/preprocess.py
+++ b/pycaret/preprocess.py
@@ -7,7 +7,7 @@
 
 import pandas as pd
 import numpy as np
-import ipywidgets as wg 
+import ipywidgets as wg
 from IPython.display import display
 from ipywidgets import Layout
 from sklearn.base import BaseEstimator , TransformerMixin
@@ -34,7 +34,7 @@ from sklearn.ensemble import RandomForestClassifier as rfc
 from sklearn.ensemble import RandomForestRegressor as rfr
 from lightgbm import LGBMClassifier as lgbmc
 from lightgbm import LGBMRegressor as lgbmr
-import sys 
+import sys
 from sklearn.pipeline import Pipeline
 from sklearn import metrics
 import datefinder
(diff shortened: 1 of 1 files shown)


Yard1:
@pycaret Since this is a preprocess change, could you look it over as well? It seems good to me, though.

pycaret/preprocess.py
```diff
@@ -187,19 +188,26 @@ def fit(self,dataset,y=None): # learning data types of all the columns
...
         data[i]=data[i].apply(str)
       except:
         data[i]=dataset[i].apply(str)
 
+    # Fix the conversion of numerical features to float and prevent some raising errors
+    # If the object column contains numbers separate by commas they will replaced by 'point'
+    # If there is no commas it will do nothing
     for i in self.numerical_features:
+      if data[i].dtype == 'object':
+          data[i] = data[i].str.replace(',', '.')
```
Yard1, line 203: Can you also add `.strip()`?

pycaret: @Yard1 It's not a high priority so let's sit on this for now. Normally `CSV` files do not have ",". I can't think of a scenario immediately where this may become problematic as PyCaret will start forcing to replace "," with "." and may result in unwanted results. 

What happens when we replace "," with ".". For example, 123,456,789 becomes 123.456.789 and when we coerced 123.456.789 into float64, is it gonna convert into 123456789 or 123.456789?

FaridAhamadaGunners: After some thought ... I think the best thing it's to always raise the error and let the user correct his error because : 
- Case 1 : we have an object column with some numbers [2.34, 3445.2, numpy.nan, 12, numpy.nan...] when we use the astype() function it raise the error that he can't convert string to float. In order to automate this task i proposed to coerce the error and convert the number to float and let the numpy.nan to nan. 

- Case 2 : We have an object column with some numbers like [1234.23.23, 123.2, 23.222.1, numpy.nan...]. If we use the astype() function it's the same result as the case 1. because 123.122.2334 it's not considered like a number. And if we coerce the error it convert it to numpy.nan so it's dangerous because we can lose important information. 

Conclusion : In the case 1 and all sub-case correlated to that. The numeric function of pandas works better. But with the case 2 and all case like that the numeric function could be a problem. So don't take into account my changes. 
Thanks

pycaret: I agree. Closing the PR.

Thanks for working on this anyways. Please feel free to work on open issues. Happy to have your contribution.

PR #565, opened 2020-08-29
Refine type hints

MattiaVerticchio:
I’ll work on checking parameter types against hints at runtime, to cut a lot of boilerplate code in the `setup()` function.
This PR adds support for `NoneType` (`Optional[type]`) and `Union[bool, str]` to `gpu_param`.


Yard1:
I added some optionals on my fork as well. Could you wait until I merge it, so we don't end up doing the same things? Thanks for this, though.

MattiaVerticchio:
Sure

Yard1:
Some types that you marked as `str` can in fact be a `str` or `object` (aka `Any`). Please double check.

PR #576, opened 2020-08-31
Add appropriate reference to Max Kuhn's caret

rhjp:
Seems a bit wrong to miss out the key inspiration of this package, therefore adding a small reference to it. We could also add a reference to https://github.com/tidymodels, however it seems a bit irrelevant as pycaret is only indirectly related to that.


pycaret:
@rhjp For some reason, this is failing. Can you please resubmit the PR. I am closing this one.

PR #580, opened 2020-09-01
Simplifying parameter checks and fixing type hints

MattiaVerticchio:
I set a single loop for all the boolean parameters to check via `isinstance(parameter, type)` and floats in range 0-1.
Also, I fixed the missing `Dict` and `Optional` imports from `typing`.

### DIFF
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -552,13 +552,18 @@ def setup(
 
     logger.info("Checking Exceptions")
 
-    # checking data type
-    if hasattr(data, "shape") is False:
+    # Checking data type
+    if not isinstance(data, pd.DataFrame):
         raise TypeError("data passed must be of type pandas.DataFrame")
 
-    # checking train size parameter
-    if type(train_size) is not float:
-        raise TypeError("train_size parameter only accepts float value.")
+    # Checking float parameters in range 0-1
+    floats_in_0_1 = ['train_size', 'rare_level_threshold', 'outliers_threshold',
+                     'multicollinearity_threshold', 'polynomial_threshold',
+                     'feature_selection_threshold', 'interaction_threshold']
+    
+    for param in floats_in_0_1:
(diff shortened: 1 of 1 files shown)


pycaret/classification.py
```diff
@@ -599,17 +599,32 @@ def setup(
...
-    # checking sampling parameter
-    if type(sampling) is not bool:
-        raise TypeError("sampling parameter only accepts True or False.")
+    # Checking float parameters in range 0-1
+    floats_in_0_1 = ['train_size', 'rare_level_threshold', 'outliers_threshold',
+                     'multicollinearity_threshold', 'polynomial_threshold',
+                     'feature_selection_threshold', 'interaction_threshold']
+    
+    for param in floats_in_0_1:
+        if not isinstance(eval(param), float) or not 0 < eval(param) < 1:
```
Yard1, line 612: Is it a good idea to use `eval` here? It's not a web app, so it's not probably a huge security risk, but it doesn't sit well with me. You could just do `locals()['param']` instead. `locals()` and `globals()` return dictionaries of currently set local and global variables in the form `name : value`.

pycaret/classification.py
```diff
@@ -910,61 +837,15 @@ def setup(
...
-            "remove_perfect_collinearity parameter only accepts True or False."
-        )
-
-    # html
-    if type(html) is not bool:
-        raise TypeError("html parameter only accepts True or False.")
-
     # use_gpu
-    if use_gpu != "Force" and type(use_gpu) is not bool:
+    if use_gpu is not "Force" and not isinstance(use_gpu, bool):
```
Yard1, line 846: Identity comparison (is) serves a different purpose than value comparison (==). The latter should be used for strings.

Yard1 (changes requested):

pycaret:
@Yard1 Do we still need this PR? can we close it?

PR #582, opened 2020-09-01
modify datasets.py

Ijustwantyouhappy:
# PULL REQUEST TEMPLATE

Work in progress. To be updated shortly.

### DIFF
--- a/pycaret/datasets.py
+++ b/pycaret/datasets.py
@@ -2,7 +2,7 @@
 # Author: Moez Ali <moez.ali@queensu.ca>
 # License: MIT
 
-def get_data(dataset, save_copy=False, profile=False, verbose=True):
+def get_data(dataset, save_copy=False, profile=False, verbose=False):
     
     """
     This function loads sample datasets that are available in the pycaret git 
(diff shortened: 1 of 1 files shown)


pycaret:
@Ijustwantyouhappy Can you explain the change a little bit? Was it a bug, if not, what was enhanced?

Ijustwantyouhappy:
When I use `get_data()` to fetch data from github repository, it might fail sometimes and throw some exceptions about HTTP, then I just wait until the same codes work again. It's a waste of time, isn't it? So I download 'datasets' folder from github repository and place it in my package path "Lib/site-packages/pycaret", by the way rename it with 'local_datasets' to avoid conflict with file 'dataset.py'. In this pr, I modify `get_data()` to fetch data from local first, if fail then try to get from url. It's convenient for a beginner to try demos from official documents, so I want to commit it.

wkuopt:
Could there be a chance that the dataset on github is changed? In that case we will be using the outdated local file.

Yard1:
A simple solution would be to have an argument to force re-download.

moezali1:

Issue #593, opened 2020-09-03
Error in Pycaret regression/predict_model 


papagala:
I used to be able to run [pycaret_deployment_streamlit](https://github.com/pycaret/pycaret-deployment-streamlit) with no problem. All of the sudden I started getting this error 

```python
pycaret/regression.py", line 7955, in predict_model     sys.exit("Pipeline not found")
```
![image](https://user-images.githubusercontent.com/6526276/92144962-1c811b80-eddd-11ea-8ca6-422ca0ed7549.png)

Reproduction steps

```bash
cd pycaret-deployment-streamlit
virtualenv virtual
source virtual/bin/activate
virtual/bin/pip install --upgrade pip
virtual/bin/pip install --force-reinstall streamlit
virtual/bin/pip install --force-reinstall pycaret==2.1.2
virtual/bin/streamlit run app.py --server.address 0.0.0.0 --server.port 8888
```

I'm running Python 3.8 and tried it in multiple OS (Ubuntu anr MacOsX).

**Note: I originally opened this [issue on pycaret-deployment-streamlit](https://github.com/pycaret/pycaret-deployment-streamlit/issues/2), but I feel this problem is more general and worth expanding the scope**

papagala:
I figured this out. It turned out that the model was built with Pycaret 1.0 and I was running the streamlit app with 2.1.2. I think it would be nice to put it on the docs if not already. I can help out with that.

pycaret:
@papagala Yes. Can you please put a warning under the `predict_model` function for all modules. Please submit a PR on `master`. Following the documentation format followed for other functions. See the `warnings` section for the `compare_models` as an example.

papagala:
[PR](https://github.com/pycaret/pycaret/pull/595) is up

vireshsaini:
> I figured this out. It turned out that the model was built with Pycaret 1.0 and I was running the streamlit app with 2.1.2. I think it would be nice to put it on the docs if not already. I can help out with that.

I am also facing same problem in "predict_model
    raise ValueError("Pipeline not found")"  and i trained the model and ran on same version pycaret 2.2.3
I don't know where is the gap.
I just saved the model in .pkl file and using the same model in program for testing data set..
Please help
Thanks
Viresh

PR #595, opened 2020-09-05
Adding a warning to all modules predict_model docstring

papagala:
This PR adds a warning to alert when `predict_model` is attempting to consume a model built with a different version of Pycaret than the one used to instantiate predict_model. 

Related issue [here](https://github.com/pycaret/pycaret/issues/593).

### DIFF
--- a/pycaret/anomaly.py
+++ b/pycaret/anomaly.py
@@ -3631,9 +3631,10 @@ def predict_model(model,
     
     Warnings
     --------
-    - `predict_model(model = my_model, data = my_data)` will throw an error if the model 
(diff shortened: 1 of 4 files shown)


pycaret:
@papagala Can you reword warning to this:

```
The behavior of the predict_model is changed in 2.1 without backward compatibility. As such the pipelines trained using the older version of PyCaret (<= 2.0), may not work for inference with version >= 2.1.
```

Thanks.

papagala:
@pycaret I agree with your suggestions and have incorporated them.

pycaret:
@Yard1 I am merging this in the `master` branch. Can you please also take this in `refactor-2.2`.

Thanks

Issue #492, opened 2020-08-20
New function create_webservice


pycaret:
Create a new function called create_webservice and use FastAPI to deploy web service on localhost. 

FastAPI will be optional dependency of PyCaret. 

## Who can work on this feature?

g0lemXIV:
I could start doing it at the beginning of September (I'm going on vacation next week), now I'm creating my own API, base on FastApi. We could talk about it more, and establish endpoints and schemes for that, in September.

Yard1:
@g0lemXIV What's the progress on this?

g0lemXIV:
@Yard1 I've just back from the holiday. I will start implementing web service this week. 

If I understand correctly, the `create_webservice` function in the basic version should take model, model_name, port number (optional), and path (optional). If the user trigger the function, fast API initializes REST API with POST endpoint and return prediction in the body.

In the next version, I can add authentication, counter, and database for prediction (optional) which can be useful for monitoring models.

I'm not sure how model ensembling works in pycaret, and how to implement it in rest API. But we could think about it in the next release.

pycaret:
@g0lemXIV Your understanding is 100% correct. plus obviously incorporate the dependencies in `requirements.txt` as well. We will keep `FastAPI` optional / commented in `requirements.txt` and in the docstring of `create_webservice` we can put a warning and an exception check on top. See exception checking for `interpret_model` where I have done a similar check to detect `shap` before executing the code.

If you can think of a better name then `create_webservice` I will be happy to hear that. If not, let's proceed with `create_webservice` for now.

pycaret:
@g0lemXIV All okay?

g0lemXIV:
@pycaret, yes sorry that I didn't respond. I'm creating it right now, it should be implemented in all main files, yes? (classification, regression, etc.)

g0lemXIV:
I missed it. I think, will pull request the first version on Sunday/Saturday. Should I add Pydantic to check the input data format?

g0lemXIV:
@pycaret  @Yard1 Tomorrow I will pull request the version with the web service. I did all the functionality except tests, docs, and requirements completion. Basic structure looks this:
`create_webservice(model, model_name, port=8080, api_key=True,
                      use_pydatic=True, pydantic_payload=None)`

api_key is responsible for generating a basic security level;
pydatic is responsible for the proper structure on input and output of the api.

pycaret:
[...]

PR #603, opened 2020-09-06
#492 New function create_webservice

g0lemXIV:
# New function create_webservice

The function will create a rest API for the pycaret model. Tested with a new test in test_classification. I had trouble with the proper print label of the juice dataset, can You check that, it seems like now not label as a string but integer are returning. The function returns dictionary with FastAPI main class and security key. It has pydantic automatically model generator, test on a cold start, prediction pydantic model, and exposed cors. All should parameters should be visible in /docs as swager.

EDIT: Sorry about so many conflicts, I overlooked that I'm working on a second computer without upstream branches update.


pycaret:
@g0lemXIV Hi. Thanks for working on this. There are a lot of conflicts with other files as well. Can you please submit a fresh PR to `master` only including `create_webservice` functionality otherwise I will have to manually resolve all the conflicts.

Thanks

PR #613, opened 2020-09-09
#492 Added create_webservice function to classification

g0lemXIV:
Added pydantic model to validate input and output (it can be automatically generated). Added additional security by generating api key. Added post endpoint to rest api. Added additional payload variable which is responsible for own pydantic model for input data. Function returns **dictionary** with {'key_api': fastapi class} which can be use with uvicorn, documentation should generate automatically in /docs endpoint.
Every model have own endpoint on /predict/{model_name}.

Api can be test with native fastapi testing client:
```python
from fastapi.testclient import TestClient
app = pycaret.classification.create_webservice(best, 'test_model', api_key=False)
client = TestClient(app['Not_exist'])
test_sample = json.loads(data.convert_dtypes().drop(columns=['Purchase']).loc[[0]].to_json(orient='records'))[0]
response = client.post("predict/{}".format('test_model'),json=test_sample)
```


pycaret:
@Yard1 I am merging this to the `master` now so I can start testing this. At some point, you will have to take this in `refactor-2.2`. I will let you know.

g0lemXIV:
@pycaret  I don't know am I right with this approach. Start rest API may be reached by implementing uvicorn inside create_webservice. I can change it,  but I'm afraid that can cause a problem with threat blocking. It will be better if this functionality will be in a different file after refactorization. I will think about better implementation after refactor (with async and autostart).

pycaret:
@g0lemXIV Sorry. I didn't understand your last comment. is `create_webservice` not starting rest API?

g0lemXIV:
@pycaret create_webservice returns dictionary with api_key (if set to True) and class with FastApi client. For example:
```python
app_dict = pycaret.classification.create_webservice(best, 'test_model', api_key=False)
application = app_dict[''Not_exist"]
```
In order to run it, the user should run in console 
```bash
uvicorn main:application --reload
```
or in the same file can use
```python
uvicorn.run(application)
```
but it will block threat (as I understand from documentation).

pycaret:
@g0lemXIV Sorry. I think there is confusion. I actually intended `create_webservice` as a function that will initialize the webserver behind the scene which we can use to send requests you to get the prediction out. Is it not that?

My goal at the over-all level is that the user should be able to train and deploy the model as a web service from the notebook?

I can run `create_webservice` successfully in my notebook but it returns me a class and API key but I can't send the POST request to localhost:port. 

What are your thoughts?

g0lemXIV:
> My goal at the over-all level is that the user should be able to train and deploy the model as a web service from the notebook?  

@pycaret create_webservice build a web server, not initialize it. As I said its hard to deploy a web app inside jupyter because uvicorn/fastapi block the main thread (example below)  

Zrzut ekranu 2020-09-14 123222   

I can change this to initialize the web server directly by function but it can cause some problems, with running further analysis (because threat blocking). Maybe You have any idea how to force that?

I have found a similar question on [stackoverflow](https://stackoverflow.com/questions/61577643/how-to-use-fastapi-and-uvicorn-run-without-blocking-the-thread), they were using multiprocessing but I am not sure if it is the right way.

PR #660, opened 2020-09-21
Add model code name to table display

calz1:
I've been using pycaret a lot and it is super cool!

When I was showing it to a colleague, she pointed out that it wasn't straightforward to go from the results of compare_models() over to create_model(). You had to know to look up in the documentation the various codes for the different classification or regression techniques (e.g. "Random Forest Classifier" is "rf"). This PR simply adds the codes into the table for reference so you don't need to have the documentation up all the time.

Here's a visual:
image


Yard1:
Hey, thanks for the PR. This is a very good idea. We won't merge this, as we are in the middle of a major code refactor and thus those changes would conflict, but we will implement this in the update itself. Thank you again!

pycaret:
@calz1 Thanks for working on this. As @Yard1 mentioned, we have taken the change in the refactored code hence I am closing this PR without merging it. Please feel free to work on other open issues.

PR #674, opened 2020-09-26
Added Public Bids Anomaly Detection Example

alfarias:
# Public Bids Anomaly Detection Example

Example made with Anomaly and Classification modules.


pycaret:
@alfarias Great work. Thanks for submitting.

PR #713, opened 2020-10-09
Add appropriate reference to Max Kuhn's caret

rhjp:
Seems a bit wrong to miss out the key inspiration of this package, therefore adding a small reference to it. We could also add a reference to https://github.com/tidymodels, however it seems a bit irrelevant as pycaret is only indirectly related to that.

Resubmitting from the advice here: https://github.com/pycaret/pycaret/pull/576#issuecomment-686581718


rhjp:
Fixed some new README conflicts that arose and now ready for review and merging. I'm not worried about how caret is mentioned, but as long as it is, I will sleep better.

pycaret:
@rhjp Thanks for bringing this up and I am sorry it falls between the cracks. I have now edited the README.md to include the appropriate reference to the Caret library in R. I am closing this PR now without merging.

PR #829, opened 2020-11-11
updated optimize_threshold (grid, float)

aliagowani:
Updated optimize_threshold function to increase grid range from 0.00 - 0.99 to 0.0000 - 0.9999 to align with default 4 decimal places variables. Also, made the t variable to a float so it can be accepted in the predict module.


Yard1:
Thank you for your contribution.

PR #832, opened 2020-11-13
Add CodeQL security scanning

daikikatsuragawa:
I propose to introduce CodeQL (code scanning) to this project. 
This is to prevent security problems hidden in the code.

https://securitylab.github.com/tools/codeql

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,11 +13,6 @@
 name: "CodeQL"
 
 on:
-  push:
-    branches: [ master ]
-  pull_request:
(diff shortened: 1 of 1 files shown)


Yard1:
Thanks for your contribution. This sounds like a good idea, though the added CI time is concerning. Could we make it run on cron only? So that PRs and commits are not held back.

daikikatsuragawa:
> Could we make it run on cron only? So that PRs and commits are not held back.

Thank you for your review. I fixed it. 

https://github.com/pycaret/pycaret/pull/832/commits/65b1b75a586bd3f27ff53c8c939c15007d60d831

Yard1:
Thanks.

PR #864, opened 2020-11-19
Fix issues with non-string columns in a df

Yard1:
If there are columns with non-string names in the test dataframe, things will break. This PR fixes this issue.


pycaret:
@Yard1 Oh did we not fixed this in `2.2.2` already?

Yard1:
@pycaret Whoops, looks like I forgot to merge this. My bad. I think you may want to re-release 2.2.2 after this is merged.

pycaret:
@Yard1 That's okay, let's keep this in our mind for the 2.2.3 release, I will update the release notes accordingly.

Issue #884, opened 2020-11-25
Suggestion for plots: display_format='streamlit'


batmanscode:
**Is your feature request related to a problem? Please describe.**
Plots don't get displayed in streamlit because it needs `st.write(fig)` instead of `fig.show()`

**Describe the solution you'd like**

*I've noticed this only in arules because it's what I'm using right now but this will be a problem for streamlit wherever `fig.show()` is.

In [arules](https://github.com/pycaret/pycaret/blob/master/pycaret/arules.py), [`fig.show()`](https://github.com/pycaret/pycaret/blob/e6671a32c27b7fe86d3a428c89adfc42cd37983f/pycaret/arules.py#L334) can be modified to:

```
def plot_model(model, plot="2d", scale=1, display_format='default'):
.
.
.
    if display_format=='streamlit'
        st.write(fig)
    else:
        fig.show()

```
**Describe alternatives you've considered**
Copy the plot function into my own code and change `fig.show()` to `st.write(fig)`

**Additional context**
This is only an inconvenience and definitely not a big issue. I also wonder if it's just me or some super small niche who might benefit from this. Anyway, if this sounds useful, I don't mind submitting a PR.

batmanscode:
#901

Issue #319, opened 2020-08-02
plot_model, evaluate_model and mlflow not producing any output in jupyter lab


ngupta23:
When I run the `plot_model`, `evaluate_model` and `mlflow` code cells in the NLP example notebok, they do not produce any output for me (just a blank space). I am using `jupyter lab`.

## Empty plot cells 

![image](https://user-images.githubusercontent.com/33585645/89122130-8a5ac000-d48a-11ea-8fd2-eaaab7b93398.png)

## MLFlow
**The mlflow cell has been running for a long time without any output (I did not see a separate browser window open up either for the experiments).**

![image](https://user-images.githubusercontent.com/33585645/89122139-9b0b3600-d48a-11ea-897c-79db505de285.png)

ngupta23:
# Update

`plot_model` and `evalaute_model` seems to work ok in `jupyter notebook` but not in `jupyter lab`. This seems to a `jupyter lab` specific problem.

I am still having issues running the following cell in `jupyter notebook`. It does not launch anything or produce an output.

```python
!mlflow ui
```

pycaret:
@ngupta23 **plot_model** and **evaluate_model** is a known issue for Jupyter Lab. Would you like to have a look at code and try to troubleshoot?

For the second issue, can you please open localhost:5000 on your browser, it should open up MLFlow server. You can stop/interupt the last cell of your notebook and continue working. The server should stay up. Please let me know if that helps.

Even that doesn't work you can just write "mlflow ui" in the same folder as of your notebook and it should still start up a local server on https://localhost:5000

Hope it helps.

ngupta23:
I don't have any luck running mlflow from the command line either.  I get this (below), but nothing happens even when I paste that localhost link in the browser

```
(pycaret) C:\Users\Nikhil\pycaret>mlflow ui
c:\users\nikhil\.conda\envs\pycaret\lib\site-packages\waitress\adjustments.py:445: DeprecationWarning: In future versions of Waitress clear_untrusted_proxy_headers will be set to True by default. You may opt-out by setting this value to False, or opt-in explicitly by setting this to True.
  DeprecationWarning,
Serving on http://Nikhil:5000
```

pycaret:
@ngupta23 Were you able to resolve this?

bolfaine:
I still have issue with model_plot in Jupyterlab, but works in jubyter notebook. Maybe some with ipy
widgets?

ngupta23:
> 
> 
[...]

PR #901, opened 2020-12-01
display association rules plots in streamlit

batmanscode:
# issue #884

### Changes

Added `display_format`

```
def plot_model(model, plot="2d", scale=1, display_format=None):
.
.
.
    if display_format=='streamlit'
        st.write(fig)
    else:
        fig.show()

```

### Notes

* Didn't include `import streamlit as st` since this will only work inside streamlit anyway
* Didn't use [`st.plotly_chart`](https://docs.streamlit.io/en/stable/api.html#streamlit.plotly_chart) since [`st.write`](https://docs.streamlit.io/en/stable/api.html#streamlit.write) can plot many types of charts and won't have to be changed if plotly is dropped later

# Update

* Added `display_format` to anywhere `plot_model` can be called 😃
* Error handling

### Additional:
In [nlp.py](https://github.com/pycaret/pycaret/blob/plot-in-streamlit/pycaret/nlp.py) where `fig.iplot()` is used, I had to change [`asFigure = save_param`](https://github.com/pycaret/pycaret/blob/2d528733269dcb353fe8d88f0c90f02c6bc8c593/pycaret/nlp.py#L2101) to `asFigure = True` because streamlit needs a plotly object to interpret (https://discuss.streamlit.io/t/cufflinks-in-streamlit/2232/2)

```
.
.
.
if display_format=='streamlit':
    fig = df.iplot(asFigure=True) # plotly obj needs to be returned for streamlit to interpret
    st.write(fig)
else:
    df.iplot()
```

### DIFF
--- a/pycaret/arules.py
+++ b/pycaret/arules.py
@@ -255,6 +255,30 @@ def plot_model(
         None
         
     """
+    
+    
+    # error handling
+    
+    # check if model is a pandas dataframe
+    if isinstance(model, pd.DataFrame) == False:
+        raise TypeError("Model needs to be a pandas.DataFrame object.")
+        
+    # check plot parameter
+    plot_types = ["2d", "3d"]
+    
+    if plot not in plot_types:
+        raise ValueError("Plots can only be \"2d\" or \"3d\".")
+        
+    # checking display_format parameter
+    plot_formats = [None, "streamlit"]
+    
+    if display_format not in plot_formats:
+        raise ValueError("display_format can only be None or \"streamlit\".")
+        
+    """
+    error handling ends here
+    """
+   
 
     # loading libraries
     import numpy as np
--- a/pycaret/arules.py
+++ b/pycaret/arules.py
@@ -273,7 +273,7 @@ def plot_model(
     plot_formats = [None, "streamlit"]
     
     if display_format not in plot_formats:
-        raise ValueError("display_format can only be None or \"streamlit\".")
+        raise ValueError("display_format can only be None or 'streamlit'.")
         
     """
     error handling ends here
--- a/pycaret/arules.py
+++ b/pycaret/arules.py
@@ -10,22 +10,22 @@ def setup(data, transaction_id, item_id, ignore_items=None, session_id=None):
     """
     This function initializes the environment in pycaret. setup() must called before
     executing any other function in pycaret. It takes three mandatory parameters:
-    (i) data, (ii) transaction_id param identifying basket and (iii) item_id param 
-    used to create rules. These three params are normally found in any transactional 
-    dataset. pycaret will internally convert the pandas.DataFrame into a sparse matrix 
+    (i) data, (ii) transaction_id param identifying basket and (iii) item_id param
+    used to create rules. These three params are normally found in any transactional
+    dataset. pycaret will internally convert the pandas.DataFrame into a sparse matrix
     which is required for association rules mining.
 
-    
+
     Example
     -------
     >>> from pycaret.datasets import get_data
     >>> data = get_data('france')
     >>> from pycaret.arules import *
     >>> exp = setup(data = data, transaction_id = 'InvoiceNo', item_id = 'Description')
-        
+
 
     data: pandas.DataFrame
-        Shape (n_samples, n_features) where n_samples is the number of samples and 
+        Shape (n_samples, n_features) where n_samples is the number of samples and
         n_features is the number of features.
 
 
@@ -36,20 +36,20 @@ def setup(data, transaction_id, item_id, ignore_items=None, session_id=None):
     item_id: str
         Name of column used for creation of rules. Normally, this will be the variable of
         interest.
-    
+
 
     ignore_items: list, default = None
         List of strings to be ignored when considering rule mining.
 
 
     session_id: int, default = None
-        If None, a random seed is generated and returned in the Information grid. The 
-        unique number is then distributed as a seed in all functions used during the 
+        If None, a random seed is generated and returned in the Information grid. The
+        unique number is then distributed as a seed in all functions used during the
         experiment. This can be used for later reproducibility of the entire experiment.
 
 
     Returns:
-        Global variables. 
+        Global variables.
 
     """
 
@@ -118,8 +118,8 @@ def setup(data, transaction_id, item_id, ignore_items=None, session_id=None):
 def create_model(metric="confidence", threshold=0.5, min_support=0.05, round=4):
 
     """
-    This function creates an association rules model using data and identifiers 
-    passed at setup stage. This function internally transforms the data for 
+    This function creates an association rules model using data and identifiers
+    passed at setup stage. This function internally transforms the data for
     association rule mining.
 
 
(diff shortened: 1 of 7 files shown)


Yard1:
Hey @batmanscode, thanks for your contribution! I am not familiar with streamlit, but can it be applied to other plots too - for example, in regression or clustering?

batmanscode:
Hi @Yard1, yes absolutely! Adding a conditional `st.write(fig)` anywhere there is a `fig.show()` will work.

I would like to add `display_format='streamlit'` to the plots in other modules as well but it seems for some the change will have to be in `..internal/tabular` and that will take a little more time for me to go through.

And as a side note, I do recommend checking out streamlit when you have the time 😃

batmanscode:
Okay so maybe `..internal/tabular` doesn't really need it since there is an option to save the plot in `plot_model()` so a streamlit app can just display the saved image.

Yard1:
I think it would be a good idea to have a unified API. Do you think you could add that? If you have trouble with modifying `internal/tabular`, I am happy to help.

batmanscode:
Yeah I think you're right. Makes more sense to have it uniform.

Yes, I should be able to! Thank you, I will let you know if I run into any problems 😃

batmanscode:
Hi @Yard1, my last commit is failing some tests but I'm not sure why. Can you please have a look when you have some time?

batmanscode:
Btw, for plots like on [this](https://github.com/pycaret/pycaret/blob/2d528733269dcb353fe8d88f0c90f02c6bc8c593/pycaret/nlp.py#L2092) line:

image

Why is it `df3 = df2.iplot()`? I thought it should be `df2.iplot()` or `iplot(df3)`

batmanscode:
> Btw, for plots like on [this](https://github.com/pycaret/pycaret/blob/2d528733269dcb353fe8d88f0c90f02c6bc8c593/pycaret/nlp.py#L2092) line:
> 
> image
> 
> Why is it `df3 = df2.iplot()`? Based on some tutorials I saw, I thought it should be `df2.iplot()` or `iplot(df3)`. Or have I missed something? 😅

I wonder if this could be related to #319

Yard1:
@batmanscode Apologies, I must have missed this. I'll take a look as soon as I am able.

batmanscode:
No problem at all @Yard1, I just figured you must have been busy.

Yard1:
@batmanscode Could you merge master into this branch, so the tests can pass? Thanks.

Yard1:
@batmanscode So the issue was that `evaluate_models` needs to be updated any time there's a new parameter added to `plot_models`. One more thing before I merge this, though - could the default value for `display_format` be `"matplotlib"`, and could you make it throw an exception if neither `"matplotlib"` nor `"streamlit"` is passed as the `display_format` parameter? It's easier to understand than just `None`. Thanks.

batmanscode:
Oh right, thank you for the fix and explanation 😃 will keep that in mind for any future changes.

Okay sure, that does sound clearer. I will go through and add this exception. And I think you mean `"plotly"` 😆

I do have a concern about naming the default plots:
Right now `plot_model` uses plotly for every module and in this case `plotly` being default for `display_format` makes sense but, if in a future release, some modules use something other than plotly, it might get a little weird.

Or do you think that it's unlikely and `plot_model` will always use one plotting library in every module?

**For example:**
If the nlp module uses some other library for plotting, then its `plot_model` would have `display_format=newlib` or `streamlit` while all the other modules would have `display_format=plotly` or `streamlit` which would mean all the `display_format` options won't work universally.

In this case, another option could be to change `display_format` to something like `special_formatting=None` or `streamlit`. This way each module can have its own default plotting library without causing any confusion.

Yard1:
> For example:
If the nlp module uses some other library for plotting, then its plot_model would have display_format=newlib or streamlit while all the other modules would have display_format=plotly or streamlit which would mean all the display_format options won't work universally.

> In this case, another option could be to change display_format to something like special_formatting=None or streamlit. This way each module can have its own default plotting library without causing any confusion.

Fair point. Let's keep `None` for now, then. Just please make sure that a user cannot pass an invalid value.

batmanscode:
Alright will do. I think the if/else statements need to be changed as well. Right now things will work fine even if an invalid value is passed since it will just execute `fig.show()` if anything other than `'streamlit'` is given.

Right now it's:

```
if display_format == 'streamlit':
    st.write(fig)
else:
    fig.show()
```

But something like this is probably better:

```
if display_format == 'streamlit':
    st.write(fig)
if disploy_format == None:
    fig.show()
```

Yard1:
@batmanscode It's fine to leave it as is, considering we'll have value checks beforehand.

batmanscode:
Ah yes, thanks!

batmanscode:
All done @Yard1. Sorry, forgot to comment that I was finished.

pycaret/internal/tabular.py
```diff
@@ -5786,6 +5790,12 @@ def is_tree(e):
         raise TypeError(
             "feature parameter must be string containing column name of dataset."
         )
+        
+    # checking display_format parameter
+    plot_formats = [None, "streamlit"]
+    
+    if display_format not in plot_formats:
+        raise ValueError("display_format can only be None or \"streamlit\".")
```
Yard1, line 5798: Please use single quotes `'` instead of escaping double quotes

Yard1 (changes requested):
Just one thing, please change this in all strings

pycaret/internal/tabular.py
batmanscode: Done. Not sure why it's not passing all the checks though.

pycaret/arules.py
```diff
@@ -247,11 +247,38 @@ def plot_model(
...
+    
+    # check if model is a pandas dataframe
+    if isinstance(model, pd.DataFrame) == False:
+        raise TypeError("Model needs to be a pandas.DataFrame object.")
+        
+    # check plot parameter
+    plot_types = ["2d", "3d"]
+    
+    if plot not in plot_types:
+        raise ValueError("Plots can only be \"2d\" or \"3d\".")
```
Yard1, line 270: Quotes

Yard1 (changes requested):
One last thing - can you run black on all modified files? Thanks.

pycaret/arules.py
batmanscode: Sorry, I missed this.

batmanscode:
> One last thing - can you run black on all modified files? Thanks.

Done 👍🏻

Yard1:
Thanks, looks great.

Yard1:
@batmanscode Hey, are you on PyCaret slack? Could you join if you haven't already and DM me there? I have no experience with streamlit but would like to expand the support for it to yellowbrick visualizers and I could use your help. Thanks.

batmanscode:
Thank you @Yard1, I've learnt a lot doing this 😃

Sure, I can join and contact you there, happy to help. I actually haven't used slack in a few years lol.

nataliaasaa:
Hey! Does anyone knows a way to show evaluate_model results on streamlit? Thanksss

batmanscode:
> Hey! Does anyone knows a way to show evaluate_model results on streamlit? Thanksss

Hi, `evaluate_model()` creates a little UI in a notebook using ipywidgets (I think) so this will not work in a streamlit app, at least right now

You can create your own `evaluate_model()` like function/script to work in streamlit using `plot_model()` (which is what `evaluate_model()` uses internally as well)

Something like this should work 😊:

```
"""
# Evaluate Model
"""

plots = st.selectbox('select plot', ("auc", "threshold")) # docs will have the other plots, not all are supported

plot_model(model, plot=plots, display_format="streamlit")
```

PR #983, opened 2021-01-07
DRY module imports

pakallis:
- Remove duplication when importing necessary packages.
- Catch ImportError exception to avoid catching the generic exception
which might cause various side effects

### DIFF
--- a/pycaret/internal/tabular.py
+++ b/pycaret/internal/tabular.py
@@ -5,6 +5,7 @@
 # Last modified : 26/08/2020
 
 from enum import Enum, auto
+from importlib import import_module
 import math
 from pycaret.internal.meta_estimators import (
     PowerTransformedTargetRegressor,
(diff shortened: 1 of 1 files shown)


Yard1:
I don't believe it is possible to do it like this, I am afraid - the CI fails.

pakallis:
Hi @Yard1 and thanks for taking a look at this.

I will fix the broken tests by using a different way to import the modules.

pakallis:
@Yard1 The CI is green now

Yard1:
I'm kind of iffy about using importlib as I imagine it will cause issues with linting in editors. I need to give this some more thought. Thank you for taking the initiative, though.

pakallis:
Do you have any case in mind where importlib might cause issues? 

I think we could fix any linting issue in the future caused by importlib.

Yard1:
Nothing imported through importlib will be considered imported by linters, meaning they will raise warnings and errors about variables being undefined. I am not sure how severe that will be, though. I don't think we use those imports directly in the file. I'll check.

ngupta23:
Hi @pakallis @Yard1 , Could you confirm  if we need to keep this open, merge it or close it out without merging? Thanks!

Yard1:
I think we should close it as this isn't a normal way to import packages - I haven't seen it done like that in any project and it does mess up linting.

moezali1:

PR #987, opened 2021-01-08
Add tests for pycaret/utils.py

pakallis:
- Add tests for cases that didn't exist previously
- Remove unnecessary imports

### DIFF
--- a/pycaret/tests/test_utils.py
+++ b/pycaret/tests/test_utils.py
@@ -123,6 +123,10 @@ def test():
     assert isinstance(mape, float)
     assert mape >= 0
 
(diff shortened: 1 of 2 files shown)


pakallis:
I have included some tests for code that was untested in the `utils` module.

It would be great if anyone could point me to another module that is more important to test it and test coverage is low so that I can add more tests (in a new merge request).

Yard1:
Thanks for your contribution! Everything looks fine to me. We are currently in the (slow) process of refactoring some modules, so at this moment it is hard for me to say what should be done regarding tests. That being said, most tests are independent from the actual code changes being made so it should not be an issue for you to continue improving them in the meanwhile.

PR #1004, opened 2021-01-16
Fixed Dummify for satisfying sklearn pipeline structure

wkuopt:
# PULL REQUEST TEMPLATE

- Fixed Dummify for satisfying sklearn's pipeline structure.
- Added an unit test for testing the complete pycaret preprocessing pipeline.
- Added Preprocess_Path_One_Sklearn for testing: This is a clone of Preprocess_Path_One with modification on the pipeline for testing purposes. It will be removed when the pipeline is completely refactored.

### DIFF
--- a/pycaret/tests/test_preprocess.py
+++ b/pycaret/tests/test_preprocess.py
@@ -1,11 +1,11 @@
 import pandas as pd
 import numpy as np
-import xgboost as xgb
 import datetime
 from scipy import stats
 from sklearn.compose import ColumnTransformer
 from sklearn.model_selection import RandomizedSearchCV
 from sklearn.pipeline import Pipeline
+from sklearn.ensemble import RandomForestClassifier
 from sklearn.preprocessing import OneHotEncoder
 import pycaret.datasets
 import pycaret.internal.preprocess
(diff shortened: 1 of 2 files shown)


pycaret/tests/test_preprocess.py
```diff
@@ -104,6 +104,38 @@ def test_simple_imputer():
...
+    data_target = data[target]
+
+    # Initiate a pycaret pipeline
+    pycaret_preprocessor = pycaret.internal.preprocess.Preprocess_Path_One_Sklearn(
+        train_data=data, target_variable=target, display_types=False
+    )
+
+    # Append classifier to preprocessing pipeline. Now we have a full prediction pipeline.
+    clf = Pipeline(steps=[('preprocessor', pycaret_preprocessor),
+                          ('classifier', xgb.XGBClassifier())],
```
Yard1, line 127: Can we use something from sklearn instead? Ideally, we would have our own mock classes for all of this, though (though IIRC sklearn provides something like that, doesn't it?)

Yard1 (changes requested):

pycaret/tests/test_preprocess.py
wkuopt: You mean like using a classifier from sklearn instead of XGBoost? Yes I can do that. I think as  long as the implementation supports the sklearn interface, we can use them in a sklearn pipeline.

Yard1: Yeah, I feel that using something like RandomForest will be a better idea, since XGboost may lag behind any API changes

wkuopt: Thanks for the comment! I have changed XGBoost to random forest.

PR #1006, opened 2021-01-17
Fixed Reduce_Dimensions_For_Supervised_Path for satisfying sklearn pipeline

wkuopt:
# PULL REQUEST TEMPLATE

- Fixed Reduce_Dimensions_For_Supervised_Path for satisfying sklearn pipeline structure.
- All pca methods tested and passed, except for TSNE. This is because that TSNE does not have a transform method (REF: https://github.com/scikit-learn/scikit-learn/issues/5361)

### DIFF
--- a/pycaret/internal/preprocess.py
+++ b/pycaret/internal/preprocess.py
@@ -7,7 +7,7 @@ import numpy as np
 import ipywidgets as wg
 from IPython.display import display
 from ipywidgets import Layout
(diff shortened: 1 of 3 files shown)


Yard1:
@wkuopt CI fails, could you check? Thanks.

wkuopt:
> @wkuopt CI fails, could you check? Thanks.

Yes this was the backward compatibility issue that I mentioned the other day. I have a good idea on how to fix it. Will commit again asap.

PR #1008, opened 2021-01-18
Added low_memory and max_len support to arules.py

astrobenhart:
Added low_memory and max_len support to arules.py for large datasets or low resource compute spaces.

apriori.apriori already has support for low_memory and max_len so just adding this option to arules.py


Yard1:
Thanks for your contribution, looks great.

PR #1088, opened 2021-02-28
Separate y from the preprocessing pipeline

wkuopt:
1. Separated y from the preprocessing pipeline in PycaretExperiment.py and tested with a target remover added in the preprocessing pipeline. 
2. Also made sure the tests related to the refactored transformers (i.e., all transformers after the target remover in the pipeline, currently only the test_pca.py) passed.

This is a draft and it hasn't dealt with the encoding of y for classification problems. The proposal for the next step is to use the TransformedTargetClassifier class to wrap classifiers. @Yard1 could you take a look and see if this approach is reasonable? Thanks!

### DIFF
--- a/pycaret/internal/PycaretExperiment.py
+++ b/pycaret/internal/PycaretExperiment.py
@@ -4659,11 +4659,17 @@ class _SupervisedExperiment(_TabularExperiment):
         dtypes.final_training_columns.append(target)
         test_data = self.prep_pipe.transform(test_data)
 
-        self.X_train = train_data.drop(target, axis=1)
-        self.y_train = train_data[target]
-
-        self.X_test = test_data.drop(target, axis=1)
-        self.y_test = test_data[target]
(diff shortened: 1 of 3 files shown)


Yard1:
I think it would be simpler to separate the data into X and y before passing it to the pipeline, thus removing the need for a transformer that removes the target. What do you think?

wkuopt:
Yes I forgot to say that the target remover will be removed once the refactoring is done. It is now there just for testing purposes. Eventually we want to pass X and y separately.

Yard1:
This is fine, then. Let me know when I can merge it.

ngupta23:
Hi @wkuopt , Is this still in work, or should we leave it open?

wkuopt:
Hi @ngupta23, let me remove this PR. It was meant to demonstrate how X is separated from y. I will create another PR that passes all the tests. Thanks for checking up on this!

PR #1114, opened 2021-03-17
tune_model + create_model updates

ngupta23:
Added tune_model capability along with updates for create model

## Type of change

## How Has This Been Tested?

Unit tests have been added

NA

## Checklist:

### DIFF
--- a/pycaret/internal/PycaretExperiment.py
+++ b/pycaret/internal/PycaretExperiment.py
@@ -59,15 +59,15 @@ from sklearn.base import clone
 from sklearn.exceptions import NotFittedError
 from sklearn.compose import TransformedTargetRegressor
 from sklearn.preprocessing import LabelEncoder
-from sklearn.model_selection import train_test_split
+from sklearn.model_selection import train_test_split, BaseCrossValidator
 from typing import List, Tuple, Any, Union, Optional, Dict
 import warnings
 from IPython.utils import io
-import traceback
 from unittest.mock import patch
 import plotly.express as px
 import plotly.graph_objects as go
 import scikitplot as skplt
+
 
 warnings.filterwarnings("ignore")
 LOGGER = get_logger()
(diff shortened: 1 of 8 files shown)


pycaret/internal/PycaretExperiment.py
```diff
@@ -16667,27 +16634,437 @@ def tune_model(
...
+        display.display_monitor()
+
+        """
+        MONITOR UPDATE ENDS
+        """
+
+        self.logger.info("Defining Hyperparameters")
+
+        # TODO: Replace with time series specific code
+        def total_combintaions_in_grid(grid):
```
Yard1, line 16865: We can use ParameterGrid from sklearn for getting the combinations, I believe it'd be the easiest way: https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.ParameterGrid.html - and it will allow for easy random search too

Yard1 (changes requested):
Looks great, just have one suggestion that will simplify the code by a fair bit :)

pycaret/time_series.py
```diff
@@ -2008,31 +2027,409 @@ def get_folds(cv, y) -> Tuple[pd.Series, pd.Series]:
...
+            scoring=scoring,
+            train=train,
+            test=test,
+            parameters=None,
+            fit_params=fit_params,
+            return_train_score=return_train_score,
+            error_score=error_score)
+            for train, test in get_folds(cv, y))
+    # raise key exceptions
+    except (TerminatedWorkerError, KeyboardInterrupt, SystemExit):
```
TremaMiguel, line 2086: I'm just wondering if an Out of Memory Error could also happen ?

TremaMiguel (approved):

pycaret/time_series.py
ngupta23: I guess it could. We can add them as we encounter them. What do you think?

pycaret/internal/PycaretExperiment.py
ngupta23: Hi @Yard1 , That is what I am using https://github.com/pycaret/pycaret/blob/568dd5365ade055398cbf3e5e0084d7fd4085d54/pycaret/time_series.py#L2435. I am not sure it has methods to compute number of combinations.

Yard1: I think it supports len. If not you should be able to calculate the number of possible combinations using some math, we probably don't need loops here

ngupta23: Do you mean something like this? https://github.com/alan-turing-institute/sktime/blob/ee7a06843a44f4aaec7582d847e36073a9ab0566/sktime/forecasting/model_selection/_tune.py#L428

Yard1: Yes

PR #1164, opened 2021-04-14
Initial additions for Random Grid Search for Time Series

ngupta23:
Added initial code for Random Grid Search for TS

## Type of change

## How Has This Been Tested?

Unit Tests have been added

NA

## Checklist:


pycaret/tests/test_time_series.py
```diff
@@ -145,3 +145,26 @@ def test_tune_model(model, load_data):
...
+    exp.setup(
+        data=load_data,
+        fold=3,
+        fh=fh,
+        fold_strategy="expandingwindow"
+    )
+
+    model_obj = exp.create_model(model)
+    tuned_model_obj = exp.tune_model(model_obj, search_algorithm="random")
+    y_pred = tuned_model_obj.predict()
```
TremaMiguel, line 166: the `predict()` method can return an empty pandas series ?

pycaret/tests/test_time_series.py
```diff
@@ -145,3 +145,26 @@ def test_tune_model(model, load_data):
...
+    assert isinstance(y_pred, pd.Series)
+
+    expected_period_index = load_data.iloc[-fh:].index
```
TremaMiguel, line 169: perhaps we can also test some behavior here, that the predictions from tune model have lower metric value than the predictions from the original model.

pycaret/time_series.py
```diff
@@ -2480,3 +2480,46 @@ def __init__(
...
+            scoring=scoring,
+            verbose=verbose,
+            error_score=error_score,
+            return_train_score=return_train_score
+        )
+        self.param_distributions = param_distributions
+        self.n_iter = n_iter
+        self.random_state = random_state
+
+    def _run_search(self, evaluate_candidates):
```
TremaMiguel, line 2517: the `evaluate_candidates` is callable?

TremaMiguel (approved):
Looks good, just added a pair of comments.

pycaret/containers/models/time_series.py
```diff
@@ -240,7 +245,10 @@ def __init__(self, globals_dict: dict) -> None:
             "degree": [1,2,3,4,5],
             "with_intercept": [True, False]
         }
-        tune_distributions = {}
+        tune_distributions = {
+            "degree": randint(low=1, high=10),
```
Yard1, line 249: It would be best if you used PyCaret's distributions (pycaret/internal/distributions.py) - feel free to expand upon them if necessary

Yard1 (changes requested):
LGTM, just one thing here - in general distributions should be left for external search libraries, for normal random/grid search just use `tune_grid` with lists

pycaret/tests/test_time_series.py
ngupta23: Yes, I will add that when the flow is working from end to end and hyperparameter grid and choice has been tuned. Just getting the skeleton setup for now so this is a very basic unit test to make sure it runs.

pycaret/time_series.py
ngupta23: Yes. its defined in the GridSearch class.

pycaret/containers/models/time_series.py
ngupta23: Hi @Yard1 , Any particular reason why? I looked at that before implementing this, but it looks like that is needed only because you are interfacing with many external GS libraries and methods. In this case, it is all internal since it needs custom argument ordering (y, X vs X, y), etc so it wont work with something like BayesGridSearchCV from skoptimize unless we implement out own version custom for TS.

Do you see any issues leaving it as is?

pycaret/tests/test_time_series.py
ngupta23: I doubt since the `fh` will expect some value. Have you encountered a situation like this?

pycaret/time_series.py
TremaMiguel: even though it is a "private" method, perhaps we could raise an error if `evaluate_candidates` is not callable

pycaret/tests/test_time_series.py
TremaMiguel: no, I was just wondering, cause in that case the first validation `isinstance(pd.Series)` will always pass

pycaret/tests/test_time_series.py
TremaMiguel: awesome

pycaret/containers/models/time_series.py
Yard1: I would want to keep it consistent so that it will be easier when we implement wrappers for search algos in the future. Distributions would be the same, after all.

pycaret/time_series.py
ngupta23: Since we are controlling that, it should be ok. In any case, it is mimicking the sklearn and sktime implementation so we should be ok.

pycaret/tests/test_time_series.py
ngupta23: Yes, I think, it is always returning a series for now in sktime (https://github.com/alan-turing-institute/sktime/blob/6e81f2c5d1848907a920e5d37184c64cb49fe743/sktime/forecasting/base/adapters/_tbats.py#L93)

pycaret/time_series.py
ngupta23: https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/model_selection/_search.py#L1625

pycaret/containers/models/time_series.py
ngupta23: @Yard1 , I have implemented the changes to use PyCaret distributions. Please have a look

Yard1 (approved):

PR #1177, opened 2021-04-19
Fixes circular imports

ngupta23:
## Related Issuse or bug
There were circular imports when ModelContainer and MetricsContainer were imported in the same file. Fixed this.
Also, cleaned up some unit tests for time_series

## Type of change

## How Has This Been Tested?

Unit Tests are passing.

## Checklist:

### DIFF
--- a/pycaret/tests/test_time_series.py
+++ b/pycaret/tests/test_time_series.py
@@ -4,49 +4,69 @@ import pytest
 
 from random import choice, uniform, randint
 from pycaret.internal.ensemble import _ENSEMBLE_METHODS
-import numpy as np
-import pandas as pd
+import numpy as np  # type: ignore
+import pandas as pd  # type: ignore
 
+from pycaret.datasets import get_data
+from pycaret.internal.pycaret_experiment import TimeSeriesExperiment
+from pycaret.containers.models.time_series import get_all_model_containers
+
+#############################
+#### Fixtures Start Here ####
+#############################
+
+# TODO: Eventually remove this and use automatically sourced model names
 _all_models = ["naive", "poly_trend", "arima", "exp_smooth", "theta" , "auto_ets", "rf_dts"]
-_parametrize_create_model = [
-    (randint(5, 10), "naive"),
-    (randint(5, 10), "poly_trend"),
-    (np.arange(1, randint(5, 10)), "arima"),
-    (np.arange(1, randint(5, 10)), "exp_smooth"),
-    (np.arange(1, randint(5, 10)), "theta"),
-    (np.arange(1, randint(5, 10)), "auto_ets"),
-    (np.arange(1, randint(5, 10)), "rf_dts"),
-]
 
 @pytest.fixture(scope="session", name="load_data")
 def load_data():
     """Load Pycaret Airline dataset."""
-    from pycaret.datasets import get_data
-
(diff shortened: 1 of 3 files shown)


pycaret/tests/test_time_series.py
```diff
@@ -59,25 +79,31 @@ def load_ts_models(load_setup):
...
-    from pycaret.internal.pycaret_experiment import TimeSeriesExperiment
+##########################
+#### Tests Start Here ####
+##########################
 
+def test_create_model(load_setup, model_parameters, load_data):
+    """test create_model functionality
+    """
     exp = TimeSeriesExperiment()
+    for name, fh in model_parameters:
```
TremaMiguel, line 94: I think that here when pytest collects the test it would not report the tests as separate

```
test_time_series.py::test_create_model[8-ARIMA]                   
test_time_series.py::test_create_model[6-Polynomial]
test_time_series.py::test_create_model[7-ETS]
```
maybe there's a way to `yield` the results from fixtures or passing the results from the fixture throught the `argvalues` argument of `pytest.mark.parametrize`. 

Some references: [here](https://docs.pytest.org/en/stable/reference.html?highlight=pytest%20mark%20parametrize#pytest.python.Metafunc.parametrize), [here](https://stackoverflow.com/questions/52764279/pytest-how-to-parametrize-a-test-with-a-list-that-is-returned-from-a-fixture) and [here](https://stackoverflow.com/questions/50231627/python-pytest-unpack-fixture/56865893)

ngupta23: Thanks @TremaMiguel , Yes, it would be nice to convert it to a parameterized test. I will experiment more with it before the next PR. Thanks for the references.

ngupta23:
I will merge this since @Yard1 already reviewed it in person.

pycaret/tests/test_time_series.py
TremaMiguel: Looks like one can pass a function to `pytest.mark.parametrize()` , for example,

```python
def model_parameters():

    globals_dict = {"seed": 0, "n_jobs_param": -1}
    model_containers = get_all_model_containers(globals_dict)

    model_names_ = []
    for model_name in model_containers.keys():
        if not model_name.startswith(("ensemble")):
            model_names_.append(model_name)

    parameters = [
        (name, np.arange(1, randint(6, 24)) if i%2==0  else randint(6, 24))
        for i, name in enumerate(model_names_)
    ]

    return parameters
```

this function can be called inside `test_create_model` 

```python
@pytest.mark.parametrize("model_parameter", model_parameters())
def test_create_model(load_setup, model_parameter, load_data):
    """test create_model functionality
    """
    exp = TimeSeriesExperiment()
    name, fh = model_parameter
    # Need to create individual setup for each model since the `fh` will be different for all models
    exp.setup(
        data=load_data, fold=3, fh=fh, fold_strategy="expandingwindow", verbose=False
    )
    model_obj = exp.create_model(name)

    y_pred = model_obj.predict()
    assert isinstance(y_pred, pd.Series)

    fh_index = fh if isinstance(fh, int) else fh[-1]
    expected_period_index = load_data.iloc[-fh_index:].index
    assert np.all(y_pred.index == expected_period_index)
```

will be adding this in my next PR.

Issue #254, opened 2020-06-29
interpret_model (reason) on an other dataset than the train/test


bourbe:
Hi,

I contruct a model with pycaret and when I launch this:

classification.interpret_model(classification_xgb, plot = 'reason', observation = 0), it gives me the interpretation on observation 0 on the test dataset

> How is it possible to do the same with a dataset that I upload in .csv for example or a dataset different of train/test

regards

pycaret:
@bourbe Not possible yet. Adding this to **enhancement** list.

pycaret:
# Open for contribution now

jbechtel:
@pycaret I'd be interested in contributing to this, however I'd like your view on its feasibility. From the SHAP side of things, it is easy. Looks like it's just 
```
explainer = shap.TreeExplainer(model)
shap_values = explainer.shap_values(X_new_sample)
```
Taking a brief look at the code, it seems like pycaret currently just grabs `X_train` or `X_test` from `global` scope, and uses that for `X_new_sample`. The problem then is tracking down what transformations `X_train` and `X_test` have already undergone. Is this as easy as applying the `prep_pipe` from `global` scope via `prep_pipe.transform(X_new_sample)`? It's unclear to me whether there are other transformations that may occur outside `prep_pipe` that are necessary to get `X_new_sample` and `X_train`/`X_test` to be compatible.

Yard1:
@jbechtel That should be all that's necessary. I'll assign this issue to you, thanks.

PR #1184, opened 2021-04-20
add X_new_sample option to interpret_model

jbechtel:
## Allows new data to be passed to `interpret_model`
  - Previously, `pycaret.internal.tabular.interpret_model` only worked for observations within the training/test data.
  - This PR adds an option `X_new_sample: Optional[pd.DataFrame] = None` which allows the user to use other data that is in the exact same format as the original dataset. 

Fixes: #254 

In order to use outside data, the new data is simply transformed by the `prep_pipe` from global scope via `test_X = prep_pipe.transform(X_new_sample)`.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)
- [x ] This change requires a documentation update

## How Has This Been Tested?

Tests plots are generated in `pycaret/tests/test_classification_plots.py` and `pycaret/tests/test_regression_plots.py` which use the first ten rows of the dataset. The plots are generated without error.

### DIFF
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -1625,6 +1625,7 @@ def interpret_model(
     feature: Optional[str] = None,
     observation: Optional[int] = None,
     use_train_data: bool = False,
+    X_new_sample: Optional[pd.DataFrame] = None,
     save: bool = False,
     **kwargs,
 ):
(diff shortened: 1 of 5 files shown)


Yard1 (approved):
Thanks, LGTM.

PR #1198, opened 2021-04-25
Ts scikit models

ngupta23:
Added more scikit models

- [x] dt_cds_dt
- [x] ridge_cds_dt
- [x] lasso_cds_dt
- [x] lar_cds_dt
- [x] llar_cds_dt
- [x] br_cds_dt
- [x] lr_cds_dt
- [x] huber_cds_dt
- [x] par_cds_dt
- [x] omp_cds_dt
- [x] en_cds_dt
- [x] gbr_cds_dt
- [x] ada_cds_dt
- [x] knn_cds_dt

## Type of change

## How Has This Been Tested?

Unit tests are passing locally

None

## Checklist:


ngupta23:
Sorry, don't review for now. I will merge with @TremaMiguel 's eventual PR and then you can review it.

pycaret/containers/models/time_series.py
```diff
@@ -271,6 +276,11 @@ def __init__(self, globals_dict: dict) -> None:
         )
 
 
+######################################
+#### CLASSICAL STATISTICAL MODELS ####
+######################################
+
+
```
TremaMiguel, line 283: To make more clear and clean the `__init__` of the ArimaContainer, is it possible to decompose this process into methods that get called in the `__init__` ?. For example `return_order_related_params` can be a method of this class called during init 

```python
class ArimaContainer(TimeSeriesContainer):

    def __init__():
          ....
         self.return_order_related_params()

    def return_order_related_params(self):

```

pycaret/containers/models/time_series.py
```diff
@@ -562,26 +572,114 @@ def __init__(self, globals_dict: dict) -> None:
...
+        sp = sp if sp is not None else 1
+        tune_grid = {
+            "sp": [sp],
+            "deseasonal_model": ["additive"],
+            "degree": [1],
+            "window_length": [10],
+            "regressor__fit_intercept": [True, False],
+            "regressor__normalize": [True, False],
+        }
+        tune_distributions = {
```
TremaMiguel, line 625: Same here, it is not easy to follow the `__init__` method, maybe for all classes could go methods like `get_tune_grid`  and `get_tune_distributions`

TremaMiguel: maybe these methods `get_tune_grid` and `get_tune_distributions` can be a part of the parent class `TimeSeriesExperiment`, we can actually force each container to have this method trough an `abstractmethod`.

ngupta23: I like the idea. Currently, I was just following the approach that already existed in other modules, but this modularization is definitely something that I would like to have as well. Since it can potentially be leveraged by other tasks as well (regression, classification, etc.), lets quickly discuss with @Yard1 before making this change. I will open a Issue for this so we keep track of it,

TremaMiguel: great, thanks Nikhil.

Issue #1246, opened 2021-05-13
[BUG] time_series Prediciton from output of compare models does not work


ngupta23:
**Describe the bug**
Time Series models from compare_models can not be used to predict (give error). This is not the case once they are tuned (this works).

**To Reproduce**

```python
best_baseline_models = exp.compare_models(sort='smape', n_select=3)
y_pred = best_baseline_models[0].predict()
```

```python
ValueError: The forecasting horizon `fh` must be passed either to `fit` or `predict`, but was found in neither.
```

```python
# Works after tuning
best_tuned_models = exp.tune_model(best_baseline_models[0], search_algorithm="random")
y_pred = best_tuned_models.predict()
plot_series(y_train, y_test, y_pred, labels=["y_train", "y_test", "y_pred"])
round(smape_loss(y_pred, y_test), 4)
```

**Expected behavior**
If looks like somehow `fh` has not been set in the model that is returned by compare_model.

ngupta23:
fixed in https://github.com/pycaret/pycaret/pull/1585/commits/46a2ac45439591ce7b306223a2c3a85149c36610

Issue #1245, opened 2021-05-13
[BUG] time_series "ValueError: Multiplicative seasonality is not appropriate for zero and negative values"


ngupta23:
**Describe the bug**
When y values are negative, multiplicative seasonality is not valid. 

**To Reproduce**
`uschange` data from `sktime` can be used to test this

**Expected behavior**
Don't try multiplicative seasonality in `tune_model` when `y` has negative values

PR #1247, opened 2021-05-13
Ts datasets

ngupta23:
Added more datasets

* lynx (yearly)
* uschange (quarterly)

Removed check for index (uschange does not have a period index, but still works).

## Type of change

## How Has This Been Tested?

Notebook added

Filed 2 bugs found when testing #1245 and #1246

## Checklist:


Yard1 (approved):

pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -357,6 +363,14 @@ def setup(
             )
         self.fh = fh
 
+        allowed_index_types = pd.core.indexes.period.PeriodIndex
```
Yard1, line 366: This should be a tuple in order to semantically match the variable name and show that it can be expanded in the future if needed (needs to change the f-string to support an iterable too)

Yard1 (changes requested):
Just one nitpick

pycaret/internal/pycaret_experiment/time_series_experiment.py
ngupta23: 'black' is removing the `tuple` unfortunately since there is only 1 element in it right now. Can we expand this later when we know what other types will be supported?

Yard1: You can force the tuple to stay there by doing `allowed_index_types = (pd.core.indexes.period.PeriodIndex,)`

ngupta23: ok, fixed now.

pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -363,14 +363,14 @@ def setup(
...
-        allowed_freq_index_types = pd.core.indexes.period.PeriodIndex
+        allowed_freq_index_types = (pd.core.indexes.period.PeriodIndex, )
         if (
             not isinstance(data.index, allowed_freq_index_types)
             and seasonal_period is None
         ):
             raise ValueError(
                 f"The index of your 'data' is of type '{type(data.index)}'. "
-                f"If the 'data' index is not of one of the following types: {allowed_freq_index_types}, "
+                f"If the 'data' index is not of one of the following types: {*allowed_freq_index_types,}, "
```
Yard1, line 373: Maybe
```suggestion
                f"If the 'data' index is not of one of the following types: {', '.join(allowed_freq_index_types)}, "
```

Yard1 (approved):
LGTM, one minor suggestion but it's up to you

PR #1265, opened 2021-05-20
Cleaning up models

ngupta23:

- [x] Added seasonal naive model
- [x] Cleaned up models by creating property methods instead of defining everything in __init__

## Type of change

## How Has This Been Tested?

Unit tests are passing locally

## Checklist:


pycaret/containers/models/time_series.py
```diff
@@ -2027,15 +1942,19 @@ def __init__(self, globals_dict: dict) -> None:
             "regressor__subsample": [0.5, 1],
             "regressor__colsample_bytree": [0.5, 1],
         }
+        return tune_grid
+
+    @property
+    def _set_tune_distributions(self) -> Dict[str, List[Any]]:
         tune_distributions = {
```
Yard1, line 1949: Question - why are LGBM tune distributions different than the ones in the regression container? I think it would be good to use the same ones - for other models as well. Also, `regressor__subsample`, `regressor__colsample_bytree` and `regressor__min_split_gain` fail because `UniformDistribution` is inclusive by default and `0` is not a valid value for them.

ngupta23: LGBM is mostly the same - right? Which parameters are different?

In some cases, I have used a smaller subset for the fixed (exhaustive) grid since there are too many combinations in regression case and using the same slows down the execution.

Also, why does the `UniformDistribution` not give an error for regression case? It is also using the value inclusive of 0.

Yard1: Ok, the commented parameters threw me off. Aside from those the ranges are indeed the same. The parameters that cause the 0-error aren't present at all in regression, though.

Yard1 (approved):

PR #1282, opened 2021-05-23
Fold Strategy Updates

ngupta23:

Previously, time series experiment only accepted a string as input for fold_strategy. With this change, it can also accept a sktime compatible cv object directly. If an object is passed, then the `fold` and `fh` parameters in setup are ignored and are extracted from the cv object directly.

## Type of change

## How Has This Been Tested?

Updated unit tests

None

## Checklist:

### DIFF
--- a/pycaret/internal/pycaret_experiment/tabular_experiment.py
+++ b/pycaret/internal/pycaret_experiment/tabular_experiment.py
@@ -31,14 +31,14 @@ import time
 import random
 import gc
 from copy import deepcopy
-from sklearn.model_selection import BaseCrossValidator
+from sklearn.model_selection import BaseCrossValidator  # type: ignore
 from typing import List, Tuple, Any, Union, Optional, Dict
 import warnings
 import traceback
 from unittest.mock import patch
 import plotly.express as px  # type: ignore
 import plotly.graph_objects as go  # type: ignore
-import scikitplot as skplt
+import scikitplot as skplt  # type: ignore
 
 
 warnings.filterwarnings("ignore")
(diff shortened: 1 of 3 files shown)


pycaret/internal/pycaret_experiment/tabular_experiment.py
```diff
@@ -906,17 +911,25 @@ def setup(
...
+                fold_strategy in possible_fold_strategy
+                or is_sklearn_cv_generator(fold_strategy)
+            ):
+                raise TypeError(
+                    f"fold_strategy parameter must be either a scikit-learn compatible CV generator object or one of {', '.join(possible_fold_strategy)}."
+                )
+        elif self._ml_usecase == MLUsecase.TIME_SERIES:
+            if not (
+                fold_strategy in possible_time_series_fold_strategies
+                or is_sklearn_cv_generator(fold_strategy)
```
TremaMiguel, line 928: `is_sklearn_cv_generator` from which module is imported ?

pycaret/tests/test_time_series_splitters.py
```diff
@@ -50,6 +50,43 @@ def test_setup_initialization(fold, fh, fold_strategy, load_data):
         assert exp_name.fold_generator.step_length == fh  # Since fh is an int
 
 
+def test_setup_pass_cv_object(load_data):
+
```
TremaMiguel, line 54: I think we could add docstrings to test to get a general idea of what we want to test.

TremaMiguel, line 54: I think we could add docstrings to tests to get a general idea of what we want to test.

TremaMiguel, line 54: I think we could add docstrings to tests to get a general idea of what we want to test.

TremaMiguel (approved):

pycaret/internal/pycaret_experiment/tabular_experiment.py
ngupta23: It is defined in `pycaret` itself. Just looks to see if there is a split method in the object.

pycaret/tests/test_time_series_splitters.py
ngupta23: Good idea. I have added it now.

PR #1286, opened 2021-05-23
Added catboost model

ngupta23:
Added catboost model

## Type of change

## How Has This Been Tested?

Unit tests should catch this new model

Catboost is too slow and consumes a lot of memory

## Checklist:

### DIFF
--- a/pycaret/containers/models/time_series.py
+++ b/pycaret/containers/models/time_series.py
@@ -12,7 +12,7 @@ from typing import Union, Dict, List, Tuple, Any, Optional
 from abc import abstractmethod
 import random
 import numpy as np  # type: ignore
-
+import logging
 
 from sktime.forecasting.base._sktime import _SktimeForecaster  # type: ignore
 from sktime.forecasting.compose import ReducedForecaster, TransformedTargetForecaster  # type: ignore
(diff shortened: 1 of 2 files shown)


pycaret/containers/models/time_series.py
```diff
@@ -1972,6 +1972,97 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+    def return_model_class(self):
+        try:
+            import catboost
+        except ImportError:
+            self.logger.warning("Couldn't import catboost.CatBoostRegressor")
+            self.active = False
+            return
+
+        catboost_version = tuple([int(x) for x in catboost.__version__.split(".")])
+        if catboost_version < (0, 23, 2):
```
TremaMiguel, line 1999: We can read the required catboost version from the `requirements-optional.txt` file, otherwise it could be hard to remember to update this part of the code when updating the requirements file.

pycaret/containers/models/time_series.py
```diff
@@ -1972,6 +1972,97 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+                0.5,
+            ],  # too low values leads to learning rate errors (Learning rate should be non-zero)
+            "regressor__depth": [1, 6, 12],
+            "regressor__n_estimators": np_list_arange(10, 100, 100, inclusive=True),
+            "regressor__random_strength": np_list_arange(
+                0.01, 0.8, 0.8, inclusive=True
+            ),
+            "regressor__l2_leaf_reg": [1, 30, 200],
+        }
+        if self.use_gpu:
```
TremaMiguel, line 2038: the `regressor__depth` parameter would be a `IntUniformDistribution` as the example below only if `use_gpu` is true ?

pycaret/containers/models/time_series.py
ngupta23: But the `requirements.txt` would reside outside the module and wont be accessible from within - correct?

pycaret/containers/models/time_series.py
ngupta23: There is an issue with the regressor parameters for catboost right now. If I add any `regressor__` parameters, it fails to recognize anything as a distribution. I added this as a placeholder but will eventually need to fix this.

Yard1:
> Catboost is too slow and consumes a lot of memory 
That's mostly due to it fitting 1000 trees by default. We can consider reducing that number

pycaret/containers/models/time_series.py
Yard1: Checking it like it's being done in here is correct, as the user may choose to install catboost separately.

Yard1 (approved):

PR #1306, opened 2021-05-28
tune_model fixes and cleanup

ngupta23:

**Tune Model changes**
- [x] Allow usage of a `custom_grid` with `tune_model`
- [x] `choose_better = True` works now for `tune_model`
- [x] cleaned up arguments that are not being used currently
- [x] changed default `search_algorithm` to `random` (previously `grid`)

**Create Model changes**
- [x] Metrics displayed in all caps now (were lower case earlier)

**Unit Tests Speedup**
- [x] changed most folds from 3 to 2
- [x] changed from expanding to sliding window to reduce training dataset 

## Type of change

## How Has This Been Tested?

Unit tests added as appropriate

## Checklist:

### DIFF
--- a/pycaret/internal/pycaret_experiment/time_series_experiment.py
+++ b/pycaret/internal/pycaret_experiment/time_series_experiment.py
@@ -62,7 +62,6 @@ class TimeSeriesExperiment(_SupervisedExperiment):
         self.variable_keys = self.variable_keys.union(
             {"fh", "seasonal_period", "seasonality_present"}
         )
-        return
 
     def _get_setup_display(self, **kwargs) -> Styler:
         # define highlight function for function grid to display
@@ -793,10 +792,7 @@ class TimeSeriesExperiment(_SupervisedExperiment):
         """
         MONITOR UPDATE ENDS
         """
-        # TODO: pass metrics dict to cross_validate_ts instead of metrics (also update cross_validate_ts to use scorers directly)
-        # Also change to dict comprehension (for efficiency)
         metrics_dict = dict([(k, v.scorer) for k, v in metrics.items()])
-        # metrics_dict = {k, v.scorer for k, v in metrics.items()}
 
         self.logger.info("Starting cross validation")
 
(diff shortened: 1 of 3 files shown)


pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -883,10 +876,7 @@ def tune_model(
         custom_grid: Optional[Union[Dict[str, list], Any]] = None,
         optimize: str = "smape",
         custom_scorer=None,
-        search_library: str = "pycaret",
```
Yard1, line 886: I wonder if it's a good idea to remove this. We will want to add bayesian search at some point. Perhaps it would be better to keep `search_library` but only allow `pycaret` as a value?

pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -1303,7 +1239,7 @@ def get_iter(x):
             return nc
 
         if search_algorithm is None:
-            search_algorithm = "grid"
+            search_algorithm = "random"  # Defaults to Random
 
         if search_algorithm == "grid":
```
Yard1, line 1244: We should raise an error if the `search_algorithm` passed is neither `grid` nor `random`.

Yard1 (changes requested):

pycaret/internal/pycaret_experiment/time_series_experiment.py
ngupta23: We can add it back later when it is needed. Right now it is redundant.

pycaret/internal/pycaret_experiment/time_series_experiment.py
ngupta23: Yes, I will add it in the next round.

Yard1:
@ngupta23 can you add `refit=False` to both `ForecastingGridSearchCV` and `ForecastingRandomizedSearchCV`? We do our own refit so they don't need to waste time doing one themselves

pycaret/internal/pycaret_experiment/time_series_experiment.py
ngupta23: @Yard1 , I have added this check.

ngupta23:
> 
> 
> @ngupta23 can you add `refit=False` to both `ForecastingGridSearchCV` and `ForecastingRandomizedSearchCV`? We do our own refit so they don't need to waste time doing one themselves

Good point. I have made `refit = False` now.

Yard1 (approved):

Issue #1323, opened 2021-06-03
[BUG] `internal.persistence` module fails to handle few exceptions and contains few typos


KrishnanSG:
**Describe the bug**

`internal.persistence` module has minor bugs and typos

* The import exception is check on awscli rather than boto3 causing incorrect log message being displayed
* AWS/Azure/GCP connection fails exception are not handled
* eg: Successful is misspelt as "succesful"

Another problem is that the except scope used is too broad, rather this should be scoped to the specific exception (ModuleNotFoundError)
```python
# Module: internal.persistence

deploy_model(...)
...
  try:
      import google.cloud
  except:
      logger.error(
          "google-cloud-storage library not found. pip install google-cloud-storage to use deploy_model function with GCP."
      )
      raise ImportError(
          "google-cloud-storage library not found. pip install google-cloud-storage to use deploy_model function with GCP."
      )

```

**To Reproduce**

**Snippet 1**
```python
from pycaret.internal.persistence import deploy_model

model = None

# since boto3 isn't present this will fail, but the deploy_model doesn't wrap boto3 instead wraps awscli (this isn't the right one) around try...except block
deploy_model(model, 'lr', platform='aws', authentication={'bucket': 'abc'})


```

**Snippet 2**
```python
from pycaret.internal.persistence import deploy_model

model = None

# 1. google.auth.exceptions.DefaultCredentialsError exception isn't handled
# 2. gcp the authentication param expects {'project': 'name', 'bucket' : 'name'} but that isn't checked or handled, leading to unexpected results when either is None.

deploy_model(model, 'lr', platform='gcp', authentication={'bucket': 'abc'})

```

**Expected behavior**

* When aws client boto3 isn't available, the exception should be handled and appropriate log message has to be displayed.
* When AWS/Azure/GCP connection fails exception is to handled

**Additional context**

**Versions**

<details>
<= 2.3.1

</details>

PR #1324, opened 2021-06-03
Fix unhandled exceptions and typos in `internal.persistence` module

KrishnanSG:
## Related Issue or bug
  - Unhandled exception in internal.persistence and typos

Fixes: #1323

* The `internal.persistence` module was checking imports for awscli. I have replaced it with boto3 which is the one we are actually supposed to check.
* Missing parameter values for authentication was not checked. I have added checks and log messages for it
* Docstring and log messages had couple of typos. I fixed those.

## Type of change

## How Has This Been Tested?

No new functional change is made, so should be aligned with the previous test cases.

#### Describe if there is any unusual behavior of your code(Write `NA` if there isn't)
NA

## Checklist:

## Screenshots

 Original           | Updated


pycaret:
@KrishnanSG Two quick comments:

- I think `awscli` is needed to configure the was credentials through the command line. I would assume when `awscli` is installed boto3 is also installed/upgraded automatically. Is that not correct? Can I configure as credentials without awscli? Probably not.
-  All the docstring changes made in persistence.py will have to replicated in classification.py / regression.py / clustering.py / anomaly.py as those are the docstrings users see. (In future to avoid confusion, we can just remove the docstrings from all the internal files).
- @Yard1 can comment on the technical details of PR.

KrishnanSG:
@pycaret 

Regarding the first point, I tested that out, on installing **awscli**, boto3 isn't installed. 

These are 2 different libraries which are independent of each other.

> Q: Can I configure AWS credentials without awscli? If yes, should also be removing awscli from requirements-optional.txt?

Yes you can configure the aws credentials without the aws cli. The way to do that would be using the environment variables or shared file (~/.aws/credentials).

A more convenient way for notebook users is to pass them as env variables.

```python
import os
os.environ["AWS_ACCESS_KEY_ID "] = str("...")
os.environ["AWS_SECRET_ACCESS_KEY "] = str("...")
``` 

Here is the [boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#environment-variables) for the same.

Yard1:
Thanks, great work. Have the docs needed any changes in the public API modules? @KrishnanSG

Yard1 (approved):

KrishnanSG:
@Yard1 

> Thanks, great work. Have the docs needed any changes in the public API modules? @KrishnanSG

No, only docstring for `internal.persistence.py deploy_model()` docstring was corrected

KrishnanSG:
@Yard1 have made few changes to public docstrings. The docs would have to be refreshed.

Yard1 (approved):

pycaret:
@KrishnanSG Build is failing. Can you check?

pycaret:
@KrishnanSG I noticed classification.py has many changes. It seems like style changes. Can you confirm why that was done?

Yard1:
@pycaret it looks to be a spacy issue (again)

KrishnanSG:
> @KrishnanSG Build is failing. Can you check?
> 
> I also noticed `classification.py` has many changes. It seems like style changes. Can you confirm why that was done?

@pycaret
As per PEP8 standards it's advisable to use 4 spaces rather than 2 spaces which was used before.

I have auto PEP8 validator, that's the reason it has occurred.

Link to PEP8 standards: https://www.python.org/dev/peps/pep-0008/#indentation

KrishnanSG:
@pycaret @Yard1 

Spacy has introduced a breaking change and that's the the reason why build is failing

Ref: https://spacy.io/usage/v3#migrating

pycaret:
@KrishnanSG I have talked to @Yard1 and we suggest reverting the style changes caused by PEP8 guidelines in `classification.py` from this PR. The right way to do that would be to create a separate PR on the `time_series` branch and apply the same styling on the entire code base instead of just `classification.py`. Let's keep this PR specific to the `deploy_fix` fix and docstring updates.

KrishnanSG:
@pycaret 
Agreed, will revert that changes and push it.

Yard1 (approved):

ngupta23:
> 
> 
> > @KrishnanSG Build is failing. Can you check?
> > I also noticed `classification.py` has many changes. It seems like style changes. Can you confirm why that was done?
> 
> @pycaret
> As per PEP8 standards it's advisable to use 4 spaces rather than 2 spaces which was used before.
> 
> I have auto PEP8 validator, that's the reason it has occurred.
> 
> Link to PEP8 standards: https://www.python.org/dev/peps/pep-0008/#indentation

Plus we are using black to auto format any new code. So should probably maintain consistency there. Maybe this needs to be documented somewhere in the developers guide.

PR #1332, opened 2021-06-06
Time Series predict_model

ngupta23:

**New**
- Added `predict_model` functionality for time series

**Fixes**
- Fixed case mismatch in compare model metrics
- Fixed naming for model: `Auto Arima` to `Auto ARIMA` 
- Removed `groups` from all functions and methods

## Type of change

## How Has This Been Tested?

Unit tests have been added and updated. 
101 notebook has been updated and rerun with predict_model

NA

## Checklist:


pycaret/internal/pycaret_experiment/supervised_experiment.py
```diff
@@ -850,9 +850,14 @@ def compare_models(
...
-                by=sort, ascending=sort_ascending
-            )
+            if self._ml_usecase != MLUsecase.TIME_SERIES:
+                master_display = master_display.sort_values(
+                    by=sort, ascending=sort_ascending
+                )
+            else:
+                master_display = master_display.sort_values(
+                    by=sort.upper(), ascending=sort_ascending
+                )
```
Yard1, line 860: What's the reason for separating it like this? Wouldn't upper work with both?

pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -2047,20 +1977,50 @@ def predict_model(
...
+        try:
+            return_vals = estimator.predict(
+                X=data, fh=fh, return_pred_int=return_pred_int, alpha=alpha
+            )
+        except NotImplementedError:
+            # Most likely prediction interval has not been implemented.
+            # Hence try by explicitly setting it to False
+            return_vals = estimator.predict(
+                X=data, fh=fh, return_pred_int=False, alpha=alpha
+            )
```
Yard1, line 2001: We should log this as a warning

Yard1 (changes requested):

pycaret/internal/pycaret_experiment/supervised_experiment.py
ngupta23: Without converting to uppercase, it does not work for time_series since the argument is lower case but the metrics data frame is upper case. It works OK for regression/classification without this change.

pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -2047,20 +1977,50 @@ def predict_model(
...
+        #     encoded_labels=True,
+        #     round=round,
+        #     verbose=verbose,
+        # )
+
+        if fh is None:
+            fh = self.fh
+
+        try:
+            return_vals = estimator.predict(
```
TremaMiguel, line 1993: which model containers do not currently implement a `predict()` method ?

pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -2047,20 +1977,50 @@ def predict_model(
...
+            # Hence try by explicitly setting it to False
+            return_vals = estimator.predict(
+                X=data, fh=fh, return_pred_int=False, alpha=alpha
+            )
+        if isinstance(return_vals, tuple):
+            # Prediction Interval is returned
+            #   First Value is a series of predictions
+            #   Second Value is a dataframe of lower and upper bounds
+            result = pd.DataFrame(return_vals[0], columns=["y_pred"])
+            result = result.join(return_vals[1])
```
TremaMiguel, line 2007: is this join based on index value ?

pycaret/tests/test_time_series.py
```diff
@@ -194,7 +194,7 @@ def test_create_model(name, fh, load_data):
     )
     model_obj = exp.create_model(name)
 
-    y_pred = model_obj.predict()
+    y_pred = exp.predict_model(model_obj)
```
TremaMiguel, line 197: I think it is good time to pass the fixtures to a `conftest.py` file and functions to other module. 

```python
for time_series_test_utils import (
    _return_model_parameters,
    _return_model_names,
    _get_seasonal_values
)
```

and 

```python
# conftest.py
import pytest

@pytest.fixture(scope="session", name="load_data")
def load_data():   
    """Load Pycaret Airline dataset."""    
    return get_data("airline")


```

pycaret/tests/test_time_series.py
```diff
@@ -358,3 +358,55 @@ def test_tune_model_raises(load_data):
...
+        fold_strategy="expanding",
+        verbose=False,
+        session_id=42,
+    )
+
+    model = exp.create_model("arima")
+
+    expected_period_index = load_data.iloc[-fh:].index
+
+    # Default prediction
```
TremaMiguel, line 385: is there a way to parametrize these test cases to report each individual case in pytest report ?

TremaMiguel (approved):

pycaret/internal/pycaret_experiment/time_series_experiment.py
ngupta23: Yes, both the prediction series and interval dataframe have the same indices.

pycaret/tests/test_time_series.py
ngupta23: Good point. I combined it with the `create_model` test so that it tests it for all models.

pycaret/internal/pycaret_experiment/time_series_experiment.py
ngupta23: I have added warning messages now. Where is this log file captured though?

pycaret/tests/test_time_series.py
ngupta23: I am not familiar with this. Could you take a stab at this in your next PR? I will be happy to learn from that and incorporate it in my future PRs.

pycaret/internal/pycaret_experiment/supervised_experiment.py
Yard1: can we do the same for both?
```suggestion
            master_display = master_display.sort_values(
                by=sort.upper(), ascending=sort_ascending
            )
```

Yard1 (approved):

PR #1341, opened 2021-06-08
added dummylogger

tvdboom:
## Related Issuse or bug
  - Enhancement: If experiment_name=None, don't create a log file.

create_logger returns a dummy logger (does nothing when __class__.info is called) if name=None. This approach leaves the rest of the code untouched.


Yard1 (approved):

pycaret:
Hi. Can we separate the system logging parameter from log_experimemt. Lets create a new parameter called system_log (default to True). The use-case for both these loggers are different.

tvdboom:
I added the system_log parameter to (I think) all the setup functions and to the docstrings. Is it automatically added to the documentation as well?

Yard1:
Yes, the docs are auto updated. Could we actually make it so that `system_log` param can take a Logger object too (will be used instead of `create_logger`)? That way all use cases will be covered.

Yard1 (approved):
Thanks, this is great

PR #1354, opened 2021-06-13
fix: time series model types available

TremaMiguel:

Major changes include: 

1. baseline, classical, linear, neighbors and tree model types available when calling `models()`.  [link](https://github.com/pycaret/pycaret/blob/36cd00a3ac3e1176072e14fbf475d3367214d817/pycaret/internal/pycaret_experiment/time_series_experiment.py#L2360)
2. Cast metrics name to upper case to ensure highlighting minimum value in  `compare_models()`. [link](https://github.com/pycaret/pycaret/blob/36cd00a3ac3e1176072e14fbf475d3367214d817/pycaret/internal/pycaret_experiment/supervised_experiment.py#L655)

## Type of change

## Checklist:

## Screenshots

highlight_fix

### DIFF
--- a/pycaret/internal/utils.py
+++ b/pycaret/internal/utils.py
@@ -33,11 +33,11 @@ class SeasonalPeriod(IntEnum):
 
 
 class TSModelTypes(Enum):
-    baseline = 'baseline'
-    classical = 'classical'
-    linear = 'linear'
-    neighbors = 'neighbors'
-    tree = 'tree'
+    BASELINE = 'baseline'
+    CLASSICAL = 'classical'
+    LINEAR = 'linear'
+    NEIGHBORS = 'neighbors'
+    TREE = 'tree'
 
 
 def id_or_display_name(metric, input_ml_usecase, target_ml_usecase):
--- a/pycaret/internal/utils.py
+++ b/pycaret/internal/utils.py
@@ -16,7 +16,7 @@ from typing import Any, List, Optional, Dict, Tuple, Union
 from sklearn import clone
 from sklearn.model_selection import KFold, StratifiedKFold, BaseCrossValidator
 from sklearn.model_selection._split import _BaseKFold
-from enum import IntEnum
+from enum import IntEnum, Enum
 import functools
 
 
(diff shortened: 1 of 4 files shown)


ngupta23 (approved):

Yard1:
@TremaMiguel Could we actually use an Enum for this instead?

TremaMiguel:
@Yard1 you mean replacing the variable `MODEL_TYPES` with an ENUM class ?, what would be the benefit here ?

ngupta23:
> 
> 
> @Yard1 you mean replacing the variable `MODEL_TYPES` with an ENUM class ?, what would be the benefit here ?

@TremaMiguel  It is useful when you have categorical variables like this (similar to the FREQUENCY values that you have). ENUM comes with several benefits like easy edits, containment tests (allowed value test), etc.

- https://stackoverflow.com/a/27844189/8925915
- https://zetcode.com/python/enum/
- https://betterprogramming.pub/take-advantage-of-the-enum-class-to-implement-enumerations-in-python-1b65b530e1d

Yard1:
To add to what @ngupta23 said, Enum would have built-in checking for whether the values are allowed, and would allow us to use Intellisense when adding new models.

TremaMiguel:
Thanks for the references @Yard1 , @ngupta23 . I'll take a look at them and make appropriate changes.

pycaret/internal/utils.py
```diff
@@ -32,6 +32,14 @@ class SeasonalPeriod(IntEnum):
     Y = 1 #year
 
 
+class TSModelTypes(Enum):
+    baseline = 'baseline'
```
ngupta23, line 36: Looks good overall. 

Sorry for being picky. General convention is to keep the first one UPPERCASE since it is a constant. If it is not too much work, could you change it to make it consistent. Thanks!

https://docs.python.org/3/library/enum.html

ngupta23 (approved):

Yard1 (approved):
I've taken the liberty of using the Enum features more, let me know if anything is unclear @TremaMiguel

TremaMiguel:
Thanks @Yard1

PR #1361, opened 2021-06-15
fix for non default optimize metric

ngupta23:
## Related Issuse or bug
Time Series tune_model does not work with non-default `optimize` argument

```python
tune_model(model, optimize = 'MAE')
```  

```bash
~\pycaret-timeseries\pycaret\pycaret\time_series.py in fit(self, y, X, **fit_params)
   2199             raise ValueError(
   2200                 f"Refit Metric: '{refit_metric}' is not available. ",
-> 2201                 f"Available Values are: {list(scorers.keys())}",
   2202             )
   2203 
ValueError: ("Refit Metric: 'smape' is not available. ", "Available Values are: ['MAE']")
```

Fixed the above mentioned issue. Refit metric was not being passed to the Grid Search instantiation earlier.

## Type of change

## How Has This Been Tested?

Unit test has been added

## Checklist:

### DIFF
--- a/pycaret/tests/test_time_series.py
+++ b/pycaret/tests/test_time_series.py
@@ -470,6 +470,23 @@ def test_tune_model_custom_folds(load_data):
     assert len(metrics2) == custom_fold + 2  # + 2 for Mean and SD
 
 
+def test_tune_model_alternate_metric(load_data):
+    """tests model selection using non default metric"""
+    exp = TimeSeriesExperiment()
+    fh = 12
+    fold = 2
+
+    exp.setup(data=load_data, fold=fold, fh=fh, fold_strategy="sliding")
+
+    model_obj = exp.create_model("naive")
+    tuned_model_obj = exp.tune_model(model_obj, optimize="MAE")
+    y_pred = exp.predict_model(tuned_model_obj)
+    assert isinstance(y_pred, pd.Series)
+
+    expected_period_index = load_data.iloc[-fh:].index
+    assert np.all(y_pred.index == expected_period_index)
+
+
 def test_tune_model_raises(load_data):
     """Tests conditions that raise an error due to lack of data"""
 
--- a/pycaret/internal/pycaret_experiment/time_series_experiment.py
+++ b/pycaret/internal/pycaret_experiment/time_series_experiment.py
@@ -17,6 +17,7 @@ import pycaret.internal.patches.sklearn
 import pycaret.internal.patches.yellowbrick
 from pycaret.internal.logging import get_logger
 from pycaret.internal.Display import Display
+
 from pycaret.internal.distributions import *
 from pycaret.internal.validation import *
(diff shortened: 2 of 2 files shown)


KrishnanSG (approved):

Yard1 (approved):

pycaret/tests/test_time_series.py
```diff
@@ -470,6 +470,23 @@ def test_tune_model_custom_folds(load_data):
     assert len(metrics2) == custom_fold + 2  # + 2 for Mean and SD
 
 
+def test_tune_model_alternate_metric(load_data):
+    """tests model selection using non default metric"""
+    exp = TimeSeriesExperiment()
+    fh = 12
+    fold = 2
+
+    exp.setup(data=load_data, fold=fold, fh=fh, fold_strategy="sliding")
```
TremaMiguel, line 479: it may be a good idea to later define a fixture for setup with `fold_strategy="sliding"`

pycaret/tests/test_time_series.py
```diff
@@ -470,6 +470,23 @@ def test_tune_model_custom_folds(load_data):
...
+
+    model_obj = exp.create_model("naive")
+    tuned_model_obj = exp.tune_model(model_obj, optimize="MAE")
```
TremaMiguel, line 482: we could parametrize this to accept `MAE, SMAPE` and all the metrics available in time series module.

pycaret/tests/test_time_series.py
```diff
@@ -470,6 +470,23 @@ def test_tune_model_custom_folds(load_data):
...
 
+def test_tune_model_alternate_metric(load_data):
+    """tests model selection using non default metric"""
+    exp = TimeSeriesExperiment()
+    fh = 12
+    fold = 2
+
+    exp.setup(data=load_data, fold=fold, fh=fh, fold_strategy="sliding")
+
+    model_obj = exp.create_model("naive")
```
TremaMiguel, line 481: is there a reason for testing the naive model? I image this makes test run faster as you only want to test non default parameters.

ngupta23: Yes, it is to make it faster. I just want to test that the flow works with non-default parameters (so 1 model would be OK). There is already a flow that tests all models through tune_model with default arguments so I think this is a good compromise,

pycaret/tests/test_time_series.py
ngupta23: I think we will have to re-architect the unit tests at some point after alpha release. For example, we could have a fixture for create models also so that could be reused in all tests instead of creating them again in each test.

pycaret/tests/test_time_series.py
ngupta23: Yes, good idea. Lets add it as an enhancement to make when we refactor the tests.

PR #1363, opened 2021-06-16
added display for test metrics in predict model

ngupta23:
* predict_model displays the test metrics now when verbose is set to True (default)
* Note that this is slightly different from the regression.classification case. There are several corner cases such as: 
  - If the model has not been finalized, prediction match y_test
  - If the model has been finalized, predictions will not match the indices from y_test (metrics display NaN in that case)
  - Also, the user can use a different fh length in predict compared to what was used in setup. In this case, only the common indices in prediction and test are used to compute the metrics.

## Type of change

## How Has This Been Tested?

Unit test has been added

#### Describe if there is any unusual behaviour of your code

Warnings are issued for all corner cases identified above

## Checklist:

### DIFF
--- a/pycaret/internal/pycaret_experiment/time_series_experiment.py
+++ b/pycaret/internal/pycaret_experiment/time_series_experiment.py
@@ -2025,6 +2025,10 @@ class TimeSeriesExperiment(_SupervisedExperiment):
         data = None  # TODO: Add back when we have support for multivariate TS
 
         X_test_ = self.X_test.copy()
+        # Some predict methods in sktime expect None (not an empty dataframe as
+        # returned by pycaret). Hence converting to None.
+        if X_test_.shape[0] == 0 or X_test_.shape[1] == 0:
+            X_test_ = None
         y_test_ = self.y_test.copy()
 
         if fh is None:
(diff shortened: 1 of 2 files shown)


Yard1:
@ngupta23 Thanks, I'll edit `create_model` and we can merge this

ngupta23:
> 
> 
> @ngupta23 Thanks, I'll edit `create_model` and we can merge this

@Yard1  What is the change needed to `create_model`?

Yard1:
@ngupta23 Displaying the frame returned by `predict_model` if `cross_validation=False`

pycaret/internal/pycaret_experiment/time_series_experiment.py
```diff
@@ -2024,9 +2024,30 @@ def predict_model(
...
 
+        display = None
+        try:
+            np.random.seed(self.seed)
+            if not display:
+                display = Display(
+                    verbose=verbose,
+                    html_param=self.html_param,
+                )
+        except:
```
TremaMiguel, line 2045: can we catch here the expected error instead of catching all kind of errors ?

TremaMiguel (approved):

pycaret/internal/pycaret_experiment/time_series_experiment.py
ngupta23: I have not delved into this part. This is copied over from the regression module. @Yard1 any thoughts on this?

Yard1: It would be a good idea, but I can't remember what sort of an error can pop up there

PR #1372, opened 2021-06-18
Update clustering.py

susmitpy:
Since clustering is an unsupervised machine learning module, in case of multicollinearity, the feature having less correlation with the target cannot be removed since there is no target variable.

## Related Issuse or bug
  - Info about Issue or bug

Fixes: #[issue number that will be closed through this PR]

Since clustering is an unsupervised machine learning module, in case of multicollinearity, the feature having less correlation with the target cannot be removed since there is no target variable.

## Type of change

Fixing documentation.


Yard1 (approved):
Thanks!

Issue #953, opened 2020-12-22
check_metric reports same value for MSE and RMSE for a regression use case


pahire:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

pahire:
Sample code and results.
unseen_data_mse = check_metric(unseen_predictions['fieldA'], unseen_predictions.Label, 'MSE')
print(unseen_data_mse)
-------------
79959.3149
-------------

unseen_data_rmse = check_metric(unseen_predictions['fieldA'], unseen_predictions.Label, 'RMSE')
print(unseen_data_rmse)
--------------
79959.3149
-------------

MSE values seems correct, while the RMSe value is not performing the square root.

Yard1:
Is it possible to share the dataset with us?

pahire:
No, the dataset is confidential.

pahire:
I see proper values for RMSE and MSE in compare_models, tune_model and ensemble_model calls. RMSE is the square root of MSE. The value of MSE is of a similar order of magnitude as seen above.

Yard1:
Would it be possible for you to check if this issue can be reproduced on a public dataset (like the ones provided by `pycaret.datasets`)? Thanks.

pahire:
I have seen it before. Would this link help ? In cells 52, 53 I encountered the same problem. The dataset and code is available there.

https://www.kaggle.com/paviah/credit-score-prediction-light-gbm-pycaret

pahire:
class RMSEMetricContainer(RegressionMetricContainer):
    def __init__(self, globals_dict: dict) -> None:

        super().__init__(
            id="rmse",
            name="RMSE",
            score_func=metrics.mean_squared_error,
            greater_is_better=False,
            args={"squared": False},
            scorer="neg_root_mean_squared_error",
        )
Isn't the score_func wrong ?

tezansahu:
Yeah, I think it should be modified to `score_func=np.sqrt(metrics.mean_squared_error)`

tezansahu:
I'll try to submit a PR with the correction.

PR #1438, opened 2021-07-08
Corrected RMSE container

tezansahu:
Initially, using `pycaret.utils.check_metric()` with RMSE gave same result as using it with MSE. Updated the RMSE container to now calculate the square root of MSE & return as the RMSE.

## Related Issuse or bug
  - Using the `check_metric()` function with `'RMSE'` yielded same result as that with `'MSE`'. In the `pycaret/containers/metrics/regression.py` file, the `score_func` for RMSE Container was incorrect (square root wasn't applied). 

Fixes: #953

Created a function `root_mean_squared_error()` in the RMSE Container class, which calculated the RMSE value. Set this function as the `score_func` for the container.

## Type of change

NA

## Checklist:

## Screenshots

 Original           | Updated
 <a href="https://imgbb.com/">Screenshot-2021-07-08-155548</a>| <a href="https://ibb.co/2nFmh1x">image</a> |

### DIFF
--- a/pycaret/containers/metrics/regression.py
+++ b/pycaret/containers/metrics/regression.py
@@ -178,11 +178,22 @@ class MSEMetricContainer(RegressionMetricContainer):
 
 class RMSEMetricContainer(RegressionMetricContainer):
     def __init__(self, globals_dict: dict) -> None:
+        def root_mean_squared_error(
+            y_true, y_pred, *, sample_weight=None, multioutput="uniform_average"
+        ):
+            return np.sqrt(
+                metrics.mean_squared_error(
+                    np.abs(y_true),
+                    np.abs(y_pred),
+                    sample_weight=sample_weight,
+                    multioutput=multioutput,
+                )
+            )
 
         super().__init__(
             id="rmse",
             name="RMSE",
-            score_func=metrics.mean_squared_error,
+            score_func=root_mean_squared_error,
             greater_is_better=False,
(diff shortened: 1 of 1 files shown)


Yard1:
Thanks for catching this. Looking at sklearn docs, `metrics.mean_squared_error` can return RMSE if `squared` is set to False. Can we do that instead?

tezansahu:
Currently, the `arg` shown in the Original Screenshot tell me that `squared` is set to False, yet it is not giving me the RMSE value. Hence, I thought it would we better to create teh separate function, similar to the case of the `RMSLEMetricContainer`

Yard1:
>>> Currently, the args shown in the Original Screenshot tell me that squared is set to False, yet it is not giving me the RMSE value. Hence, I thought it would we better to create the separate function, similar to the case of the RMSLEMetricContainer

Oh, my mistake. I didn't notice it being there. Are you sure it's not working correctly? If so, that seems to be an issue with how we apply arguments. I will take a look at it. Can you try using sklearn's `metrics.mean_squared_error` directly on some dataset with and without `squared=False`?

Yard1:
Right, I believe I found the issue. We are ignoring the arguments when calling the function directly. Thank you for taking the time to make this PR but this should be fixed in general. I'll close it and push a fix to the underlying issue.

tezansahu:
Ok. Thank you.

PR #1451, opened 2021-07-15
Add transformation

IncubatorShokuhou:
Add transformation

## Type of change

## How Has This Been Tested?

## Checklist:

## Screenshots

 Original           | Updated

### DIFF
--- a/pycaret/internal/pycaret_experiment/tabular_experiment.py
+++ b/pycaret/internal/pycaret_experiment/tabular_experiment.py
@@ -39,6 +39,7 @@ from pycaret.internal.utils import (
     get_model_name,
     normalize_custom_transformers,
 )
+
 from pycaret.internal.validation import *
 from sklearn.compose import ColumnTransformer
 from sklearn.decomposition import PCA, IncrementalPCA, KernelPCA
@@ -58,6 +59,8 @@ from sklearn.preprocessing import (
     MinMaxScaler,
     OrdinalEncoder,
     PolynomialFeatures,
+    PowerTransformer,
+    QuantileTransformer,
     RobustScaler,
     StandardScaler,
 )
@@ -382,7 +385,13 @@ class _TabularExperiment(_PyCaretExperiment):
         return
 
     def _set_up_mlflow(
-        self, functions, runtime, log_profile, profile_kwargs, log_data, display,
+        self,
+        functions,
+        runtime,
+        log_profile,
+        profile_kwargs,
+        log_data,
+        display,
     ) -> None:
         return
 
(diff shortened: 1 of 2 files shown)


IncubatorShokuhou:
@tvdboom Please review.

pycaret/internal/pycaret_experiment/tabular_experiment.py
```diff
@@ -1106,6 +1118,28 @@ def setup(
 
             self._internal_pipeline.steps.append(("normalize", normalize_estimator))
 
+        # transformation ========================================= >>
+
+        if transformation:
+            if transformation_method == "yeo-johnson":
+                transformation_estimator = PowerTransformer(
+                    method="yeo-johnson", standardize=True, copy=True
```
tvdboom, line 1126: I would suggest to set `standardize=False` and add this transfomation step before normalization. So, if the user wants to normalize after this, he can just use the `normalize` parameter (see #1453)

tvdboom: Additionally, we could add box-cox as an additional option

IncubatorShokuhou: In fact, yeo-johnson is an extension of box-cox over the real number field (see Johnson, Y. R. A. . (2000). A new family of power transformations to improve normality or symmetry. Biometrika, 87(4), 954-959. ). Therefore, I suggest that we implement the auto-selection of box-cox and yeo-johnson. When the label is all positiveuse, use box-cox to speed up the calculation, and automatically switch to yeo-johnson when the label is not all positive.

tvdboom: The transformation is applied on every column in the dataset, so we would need to check if all columns are strictly positive or make two lists of columns (one strictly positive and the other not) and add a columntransformer containing two estimators, one for each feature group. Besides, it still could fail if the user wants to apply the pipeline to new data, without realizing it contains negative values in a column.

tvdboom: In the spirit of simplicity, maybe we should leave it as is. Box-cox is almost never used (as far as I know) and this way we don't need to change the API. Only change the normalization part I mentioned in the first comment. What do you think?

IncubatorShokuhou: Yes, I totally agree with you. I didn't think about the test dataset before.

IncubatorShokuhou: I have revised it according to your first comment.

IncubatorShokuhou: BTW, there seems to be an indentation error (`transformation` et.al. should be behind `Preprocessing`). Maybe you can help review it to see if I fixed it correctly.

tvdboom: looks good. accepting the PR

tvdboom (approved):

PR #1469, opened 2021-07-23
fix: blending compatible with sktime 0.7.0

TremaMiguel:

Reimplement EnsembleForecaster with sktime>=0.7.0

## Type of change

## Checklist:

### DIFF
--- a/pycaret/internal/ensemble.py
+++ b/pycaret/internal/ensemble.py
@@ -46,7 +46,7 @@ class _EnsembleForecasterWithVoting(_HeterogenousEnsembleForecaster):
         self.forecasters = forecasters
         self.method = method
         self.weights = weights
-        super(EnsembleForecaster, self).__init__(
+        super(_EnsembleForecasterWithVoting, self).__init__(
             forecasters=self.forecasters, n_jobs=n_jobs
         )
 
--- a/pycaret/internal/ensemble.py
+++ b/pycaret/internal/ensemble.py
@@ -1,140 +1,129 @@
-# # Module: internal.ensemble
-# # Author: Miguel Trejo
-# # License: MIT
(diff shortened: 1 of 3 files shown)


pycaret/internal/ensemble.py
```diff
@@ -1,140 +1,129 @@
...
+            pred_w = np.average(pred_forecasters, axis=1, weights=self.weights)
+            return pd.DataFrame(pred_w, index=pred_forecasters.index)
+
+    @property
+    def _set_weights_none(self):
+        self.weights = None
+
+    @property
+    def _set_weights_uniform(self):
+        self.weights = np.ones(len(self.forecasters))
```
Yard1, line 129: Shouldn't those be methods? Is this how sktime does it?

TremaMiguel: Sktime original implementation doesn't handle `voting` method, here I should be changing these to appropiate setter for `@property`

```python
@property
 def weights(self):
      return self.weights

@weights.setter
def weights(self, value):
    self.weights = value
```

and then call

```python
elif self.method in self._not_required_weights and self.weights:
        warnings.warn(
            "Unused 'weights' argument. When method='mean' or method='median', 'weights' argument is not provided. Setting weights to `None`"
        )
        self.weights = None
```

TremaMiguel: @Yard1 I think we should open a PR to include `voting` in sktime original implementation, what are your thoughts on this ?

Yard1: That's a great idea @TremaMiguel

ngupta23 (approved):

Yard1 (approved):
LGTM!

Issue #1573, opened 2021-09-14
Installation error on Google Colab and local Ubuntu 20.04


souravjamwal77:
**Couldn't install the PyCaret on Google Colab and Local Ubuntu 20.04**
This is related to unmatched dependencies, as pyLDAvis 3.3.1 (latest) requires numpy>=1.20 and yellowbrick requires numpy<=1.20, which is creating a conflict for installing NumPy. I installed PyCaret on my system by downgrading pyLDAvis to 3.2.2 

**To Reproduce**
- just try to install PyCaret in a new virtualenv or on Google Colab
I tried on python 3.7 and 3.8, both.
```python
pip install pycaret
```
I ran this within virtual environment on my local but failed on NumPy version resolution.
and on Google Colab it failed on multiple packages, below are those packages

Google Colab Ouput:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
google-colab 1.0.0 requires requests~=2.23.0, but you have requests 2.26.0 which is incompatible.
datascience 0.10.6 requires folium==0.2.1, but you have folium 0.8.3 which is incompatible.
albumentations 0.1.12 requires imgaug<0.2.7,>=0.2.5, but you have imgaug 0.2.9 which is incompatible.

```

**Expected behavior**
PyCaret's dependencies are not locked as per SemVer and it is trying to download the latest version of each package whose required version is not mentioned which is causing the problem as different packages have similar dependencies with different versions.

**Additional context**

We should pin the dependencies in requirements.txt with SemVer notation like ~= which will let the bug updates to be updated but will ignore the minor and major versions.

**Versions**
<details>

pycaret version: 2.3.3
python versions: 3.7, 3.8
</details>

**Proposed Solution**
We should pin the dependencies in the _requirements.txt_ file with **SemVer notation like ~=**, so that only bug releases could update and no major and minor version could update

moezali1:
@souravjamwal77 I just tried installing on Google Colab. I got warnings but the installation works fine:

![image](https://user-images.githubusercontent.com/54699234/133264398-6a2d4cf2-78e4-4510-a344-9015e3c160d2.png)

souravjamwal77:
@moezali1, yes it's installed on Google Colab, but running on unmet dependencies still concerns me.
[...]

PR #1584, opened 2021-09-19
Pin dependencies using SemVer notation

souravjamwal77:
## Related Issue or bug
  - With this code, pycaret dependencies are pinned using SemVer (MAJOR.MINOR.BUG/FIX) notations **~=** which will let the pip or pipenv update the BUG/FIX version part but will keep the MAJOR and MINOR version locked.

Fixes: #1573 

Previously dependencies were written without any specific version so  I changed that and locked the dependencies on Major and Minor updates. So, that even if there is a new release of any new dependency with Major and Minor version changes, it wouldn't break our installation. And if there is any bug update or any fixes on that dependency, it will be allowed to update.

I've made changes in the following files:-
`
requirements.txt
requirements-test.txt
requirements-options.txt
`

Also, just to be clear, wherever a version wasn't mentioned I added the latest version there as if there is no version mentioned then pip or pipenv automatically gets the latest release. Hence, I added the latest version with SemVer notations **~=** to let it update its BUG/FIX part.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

I testes the installation on Python 3.7.10 and 3.8.11, and it was working well.
However, my tests were failing on `pycaret.tests.test_classfication_plots.py` because the `matplotlib` wasn't closing itself, so, I wasn't able to run tests after that. However, I was able to pass the initial 4-5 tests, and then it will be stuck on matplotlib window.

So, Installations is working fine, but someone has to look into the `pytests` as it wasn't working properly.

#### Describe if there is any unusual behavior of your code(Write `NA` if there isn't)
Yes, `pytest` isn't working as it is supposed to be. When I ran `pytest .`, it was getting stuck on the `pycaret.tests.test_classfication_plots.py` matplotlib window.
So, someone has to see why the matplotlib window wasn't closing in this case.

## Checklist:

## Screenshots
See issue #1573 for detailed info and screenshots.
 Original           | Updated


souravjamwal77:
@Yard1 could you please look into this? 
I've tested this on Ubuntu 20.04 with Python 3.7.11 and 3.8.10. Let me know if you have any questions?

Yard1:
This looks good, though I don't think we should be pinning everything (ray comes to mind right away). I will take a more through look later today, thanks!

Yard1:
@souravjamwal77 CI is failing as it can't find the pandas version, can you check?

souravjamwal77:
@Yard1 I will surely look into these and will get back to you.

souravjamwal77:
@Yard1 **Don't run the CI build yet** on this branch, let me check the latest pandas version for Python 3.6 and update accordingly. Also, I will check the other dependencies also for this.
Will let you know once I'm done.

souravjamwal77:
@Yard1 I found the solution by setting the range (>=, <=) of dependency versions. I will check the installation for Python 3.6, 3.7, and 3.8 and will let you know once it's complete.

PR #1793, opened 2021-11-01
improving error message

reza1615:
Improving error message when ignore column doesn't exist in the dataset. In below you can see the improved version which mentions the column name
image


reza1615:
@moezali1: If it is possible please include this PR in the version 2.3.5

moezali1:
@reza1615 How errors will be displayed if there are more than 1 feature passed in the `ignore_features` parameter?

reza1615:
@moezali1: It will show feature one by one. Just I showed the error parameter

moezali1 (approved):

Issue #1754, opened 2021-10-22
Adding Miss Classification table and plot


reza1615:
I coded a snippet that can output a miss classification error report and a chart.
If you are Ok with this output I can make a PR
for example:

``` python
from pycaret.datasets import get_data
data = get_data('credit')
from pycaret.classification import *
s = setup(data = data, target = 'default', session_id=123, silent=True)
lightgbm = create_model('lightgbm')
pred=predict_model(lightgbm,data,raw_score=True)

error_df= plot_model(lightgbm,plot='miss_classification', data=pred)
```

![Screenshot from 2021-10-22 16-44-34](https://user-images.githubusercontent.com/2202483/138520673-da6f91a6-367c-4161-8003-d734b9346344.png)

![Screenshot from 2021-10-22 16-44-52](https://user-images.githubusercontent.com/2202483/138520732-81042c86-f688-4a4e-8cd8-42da3f0a166d.png)

moezali1:
Closing as part of the 4.0 backlog cleanup — see #4177.

PyCaret 4.0 is a ground-up rewrite (OOP-only API, sklearn-native engine, Python 3.11+, sklearn ≥ 1.7) and almost all pre-4.0 issues either reference removed APIs (`setup()`, `plot_model`, `evaluate_model`, `interpret_model`, `automl`, the legacy time-series module, the Streamlit dashboard, …) or no longer reproduce on the new code path.

**If this still affects you in 4.0, please re-file as a new issue** with:

- the experiment constructor call (`Experiment(...)`),
- the operation that fails,
- the full traceback,
- your Python / sklearn / pycaret versions.

Resources for 4.0:

- Docs: https://pycaret.org/docs
- Migrate from 3.x: https://pycaret.org/docs/guides/migrating-from-3-x
- API reference: https://pycaret.org/reference
- FAQ: https://pycaret.org/docs/resources/faqs

Thanks for filing — and apologies for the long latency. The repo had drifted; we're starting clean.

PR #1802, opened 2021-11-04
miss_classified plot

reza1615:
This commit is related to the #1754
Test code:
``` python
data = get_data('credit')
s = setup(data = data, target = 'default', session_id=123, silent=True)
lightgbm = create_model('lightgbm')
pred=predict_model(lightgbm,data,raw_score=True)
miss_classified_df=plot_model(lightgbm, prediction_df=pred,target_column='default',plot='miss_classified', top=40)
```
Screenshot from 2021-11-04 13-01-49

Screenshot from 2021-11-04 12-46-10


ngupta23:
Hi @reza1615 @moezali1 Is this PR still active?

ngupta23:
Also, why do we need to specify `target` in plot_model? This can be derived from the `setup` - correct?

PR #1813, opened 2021-11-07
Add Brier score loss to classification

nicholasdomene:
I would like to propose the addition of Brier score as classification metric.
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.brier_score_loss.html

## Type of change
Adding new metric to classification models

## Checklist:

### DIFF
--- a/pycaret/containers/metrics/classification.py
+++ b/pycaret/containers/metrics/classification.py
@@ -265,6 +265,15 @@ class MCCMetricContainer(ClassificationMetricContainer):
             score_func=metrics.matthews_corrcoef,
             scorer=metrics.make_scorer(metrics.matthews_corrcoef),
         )
+        
+class BrierMetricContainer(ClassificationMetricContainer):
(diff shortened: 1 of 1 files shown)


ngupta23 (dismissed):

ngupta23 (commented):
Could you check again. The unit tests are failing.

nicholasdomene:
Multiclass test is running Brier eventhough metric is declared as is_multiclass = False. I will look deeper into this

moezali1:
@nicholasdomene I don't think so we should add this as the default metric. There are already 6 metrics.

Even without adding it as default, users can add the metric using the `add_metric` function if they would like.

```
from sklearn.metrics import brier_score_loss

s = setup(data, target)
add_metric('brier', 'brier', brier_score_loss)
...
...
```
Please feel free to open a thread on our discussion forum. If there is enough upvote to include brier as a default metric I will happy to consider this in the future releases.

PR #1960, opened 2021-12-15
Croston

### LINKED ISSUE #1623: Add crostons model to the time_series module
**Is your feature request related to a problem? Please describe.**
This is now available in sktime. 
https://www.sktime.org/en/latest/api_reference/auto_generated/sktime.forecasting.croston.Croston.html

**Describe the solution you'd like**
Include crostons in model zoo

**Describe alternatives you've considered**
None

**Additional context**
Useful for intermittent time series forecasting

AJarman:
## Related Issue or bug

  #1623 

#Info about Issue or bug

Closes #1623 - Implements this intermittent model within the time series module.

Added Croston model to model zoo, by implementing the class CrostonContainer (inherits from TimeSeriesContainer)

## Type of change

## How Has This Been Tested?

Ran the existing test suite in github actions and provided a demonstration notebook evidencing the expected model behaviours.

NA

## Checklist:

## Screenshots

 Original           | Updated


pycaret/containers/models/time_series.py
```diff
@@ -856,6 +856,69 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+        )
+        if not self.active:
+            return
+
+        # self.seasonality_present = globals_dict.get("seasonality_present")
+        # TODO: Not sure if this is required.
+        sp = globals_dict.get("seasonal_period")
+        self.sp = sp if sp is not None else 1
+
+        self.strictly_positive = globals_dict.get("strictly_positive")
```
ngupta23, line 881: Can `crostons` support negative values? If it can, then this may not be needed. We usually use it to disable things like multiplicative seasonality when there are negative values.

pycaret/containers/models/time_series.py
```diff
@@ -856,6 +856,69 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+
+        args = self._set_args
+        tune_args = self._set_tune_args
+        tune_grid = self._set_tune_grid
+        tune_distributions = self._set_tune_distributions
+        leftover_parameters_to_categorical_distributions(tune_grid, tune_distributions)
+
+        eq_function = lambda x: type(x) is Croston
```
ngupta23, line 889: This is not needed. It is only needed when you have 2 pycaret model containers with the same base model type. In this case, we only have a single crostons container. You can choose to remove this.

pycaret/containers/models/time_series.py
```diff
@@ -856,6 +856,69 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+        super().__init__(
+            id="croston",
+            name="Croston",
+            class_def=Croston,
+            args=args,
+            tune_grid=tune_grid,
+            tune_distribution=tune_distributions,
+            tune_args=tune_args,
+            is_gpu_enabled=self.gpu_imported,
+            eq_function=eq_function,  # Added to differentiate between ExponentialSmoothing and Theta which are of same parent class
```
ngupta23, line 900: You can remove `eq_function`

pycaret/containers/models/time_series.py
```diff
@@ -856,6 +856,69 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+        )
+
+    @property
+    def _set_args(self) -> Dict[str, Any]:
+        args = {"smoothing" : 0.1} # just use the default : smoothing = 0.1
```
ngupta23, line 905: I would not hard code it here. Leave it up to `sktime` to handle that. They anyway set the defaults.

pycaret/containers/models/time_series.py
```diff
@@ -856,6 +856,69 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+    @property
+    def _set_tune_grid(self) -> Dict[str, List[Any]]:
+        # TODO: Lack of documentation here, SKtime and R implementations are default 0.1
+        tune_grid = {"smoothing" : [0.01,0.05, 0.001,0.5,0.9]}
+        return tune_grid
+
+    @property
+    def _set_tune_distributions(self) -> Dict[str, List[Any]]:
+        tune_distributions = {"smoothing": UniformDistribution(
+                lower=0.01, upper=0.9, log=False
```
ngupta23, line 917: You fixed grid above looks to be log-sampled. Are you sure you want to set log=False here?

ngupta23:
@AJarman Thank you for your first PR to `pycaret`. I changed the branch from `time_series_beta` to `time_series`. We do not publish to `time_series_beta`. That is only used to make time series package releases.

pycaret/containers/models/time_series.py
```diff
@@ -856,6 +856,69 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+
+        dummy = Croston()
+        self.active:bool = self.disable_pred_int_enforcement(
+            forecaster=dummy, enforce_pi=globals_dict["enforce_pi"]
+        )
+        if not self.active:
+            return
+
+        # self.seasonality_present = globals_dict.get("seasonality_present")
+        # TODO: Not sure if this is required.
```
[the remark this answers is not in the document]
AJarman: Thanks, will do

ngupta23:
Also, can you add back the `.gitignore` file that probably got accidentally removed?

pycaret/containers/models/time_series.py
AJarman: will do. thanks

pycaret/containers/models/time_series.py
AJarman: So in this case I would just leave an empty dict @ngupta23 ?

pycaret/containers/models/time_series.py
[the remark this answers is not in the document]
AJarman: > Also, can you add back the `.gitignore` file that probably got accidentally removed?

Yes, will do, let me resolve the issues you have raised this evening and re-do the PR please. Thanks for answering my questions 👍

pycaret/containers/models/time_series.py
AJarman: will do, thanks.

ngupta23:
Also, there should not be a need to change `.github/workflows/build_publish_ts.yml`. Can you please remove any changes to this file. Thanks!

pycaret/containers/models/time_series.py
ngupta23: You can just leave out the class property `_set_args` (since it inherits from the base class). See example here: https://github.com/pycaret/pycaret/blob/7f302caad7841c997c4ea8653a39eff7da56e681/pycaret/containers/models/time_series.py#L443

pycaret/containers/models/time_series.py
AJarman: No- revised in latest commit.

pycaret/containers/models/time_series.py
AJarman: thanks, I assume it doesnt need to be passed to super/base class constructor?

pycaret/containers/models/time_series.py
AJarman: It doesn't seem to be able to use negative values (see notebook example)

pycaret/containers/models/time_series.py
AJarman: Nevermind, I have worked this out. Removed the code

.gitignore
```diff
@@ -34,3 +34,6 @@ trained_models/
 /mlruns.db-journal
 mlflow_backend.py
 demo4.py
+/.venv
+!time_series_croston.ipynb
```
ngupta23, line 38: this ipynb should not be in the gitignore. If you do not want to push it, do not add it to the commit.

pycaret/containers/models/time_series.py
ngupta23: I meant, if the data has negative values, will the model run. If it still runs with negative y values, then you do not need the `self.strictly_positive`. Anyway, I do not see you using it anywhere in the container. You can introduce some negative values in the data and see if it runs. If it does, you can remove this statement.

.gitignore
AJarman: OK, i thought it might be useful for you and @ltsaprounis  to review, but we can remove.

pycaret/containers/models/time_series.py
```diff
@@ -856,6 +856,59 @@ def _set_tune_distributions(self) -> Dict[str, List[Any]]:
...
+            tune_grid=tune_grid,
+            tune_distribution=tune_distributions,
+            is_gpu_enabled=self.gpu_imported
+        )
+
+    @property
+    def _set_tune_grid(self) -> Dict[str, List[Any]]:
+        # Lack of documentation here, SKtime and R implementations are default 0.1
+        # using : np.logspace(start=0.01,stop=1.0,endpoint=True, base=100, num=8)/100
+        smoothing_grid: List[float] = [0.01 , 0.02 , 0.039, 0.074, 0.142, 0.272, 0.521, 1.   ]
```
ngupta23, line 900: Why such specific values? Should it not be generic like `0.01, 0.03, 0.1, 0.2. 0.4, 0.6, 0.8, 1.0`?

pycaret/containers/models/time_series.py
AJarman: I have added negative values in the example notebook, however it doesn't really handle them well / work as expected.
It doesn't raise an error, but just writes new forecasts as 0.0.

ngupta23: The question that you will have to answer is what do you do when the data has negative values? Do you disable croston model or leave it enabled? If you want to leave it enabled, then there is not need to include self.strictly_positive

If you want to take some action when negative values exist (e.g. disable the model or limit the search space somehow), then you need to define this and take the appropriate action in the container.

So what do you want to do in this case?

AJarman: I think we should disable the model for time series with negative values - it isn't really relevant in the primary use case (inventory management).
if we add back in `self.strictly_positive`, then this should disable the model for data sets with negative values?

pycaret/containers/models/time_series.py
AJarman: In the line above I used a logspace to define these so that they roughly follow the probabilities in the UniformDistribution with log=True, happy to change back to the generic grid you suggested?

pycaret/containers/models/time_series.py
ngupta23: OK, sounds good. You need to set `self.strictly_positive` first as it is currently done. Then check if it is False. If it is, then set container `active = False` and return. This is similar to what is done when checking for Prediction Interval Enforcement in the same container so you can use that as a reference.

pycaret/containers/models/time_series.py
ngupta23: OK, I think the alternate generic suggestions would be better. We are not looking for precise values, just that we sample roughly equally between 0.01-0.1 and between 0.1-1.0. You can see equivalent examples in the other models.

https://github.com/pycaret/pycaret/blob/3b1ff2268d28a536f2d96930f5d2eec06f772c29/pycaret/containers/models/regression.py#L1752

pycaret/containers/models/time_series.py
AJarman: I have found that the 'strictly positive' test in setup.py is returning `False` when there are zeros in the data, (which disables the model, and it's main use is for intermittent forecasting!). 
So, whilst the model can't really handle negatives, it does need to handle zeros.
So using 'strictly_positive' doesn't really work in this case. 

The SKLearn algorithm just forecasts 0.0 if you feed it negative values, with no errors, which is what I suggest we replicate (I think the metrics will be uncompetitive enough that in 'compare models' it wouldn't end up near the top of the pile)

Probably easier to check out the example below.


```
import pandas as pd
from sktime.forecasting.croston import Croston

# Get any dataset as not really important for the demo what it is as we will transform!
y = get_data("lynx", verbose=False)
# clip the dataset and reduce the scale by the same amount to make a dataset with a lot of zeros / intermittent style
clip_number = np.quantile(y,0.84)
#clip and turn negative
y_neg= -(y.clip(lower=clip_number) - clip_number )
# init model
forecaster = Croston(smoothing=0.1)
# print training data
print(y_neg)
# fit the model
forecaster.fit(y_neg)
# create prediction
y_pred = forecaster.predict(fh=[i for i in range(5)])
print(y_pred)
```

pycaret/containers/models/time_series.py
AJarman: Implemented as suggested

pycaret/containers/models/time_series.py
ngupta23: Yes, that is related to my original comment. If the model already takes care of this, then you can leave it enabled and remove the self.strictly_positive statement.

ngupta23 (approved):

PR #1975, opened 2021-12-19
Time Series - Upper and Lower Clamp

### LINKED ISSUE #1766: Add lower limit and upper limit for time series forecasting

AJarman:
## Related Issue or bug

Info about Issue or bug

Closes #1766

Implemented upper and lower clamp(s) to predictions made in the time series models. This also applies to the prediction `y_pred` and the prediction intervals (`upper` `lower`). These are set to default `None` in `setup()` in the functional api as well as the OOP api.
If invalid values are passed, the `TypeError` is caught, warning is logged and these values are ignored in `TimeSeriesExperiment.predict()`

## Type of change

## How Has This Been Tested?

- I have written two tests:
    -  `test_setup_upper_and_lower_clamps_is_non_breaking()` to confirm that this is a non-breaking change to setup.py and that the TimeSeriesExperiment class instances are set as expected.
    - `test_clamp_is_enforced_on_predictions()`Confirms that the clamps are enforced and that errors are not raised when passing in non-numeric values. This also ignores np.nan values where prediction intervals are not supported by models.

NA

## Checklist:

## Screenshots

 Original           | Updated
|<b>original screenshot </b> (no clamp)original| <b>updated screenshot </b> <br>(lower clamp at 400 and upper clamp at 500) clamped |

### DIFF
--- a/pycaret/internal/pycaret_experiment/time_series_experiment.py
+++ b/pycaret/internal/pycaret_experiment/time_series_experiment.py
@@ -879,7 +879,9 @@ class TimeSeriesExperiment(_SupervisedExperiment):
         verbose: bool = True,
         profile: bool = False,
         profile_kwargs: Dict[str, Any] = None,
-    ):
+        lower_clamp: Optional[Union[float,int]] = None,
+        upper_clamp: Optional[Union[float,int]] = None,
+        ):
         """
         This function initializes the training environment and creates the transformation
         pipeline. Setup function must be called before executing any other function. It takes
@@ -1024,6 +1026,14 @@ class TimeSeriesExperiment(_SupervisedExperiment):
             Dictionary of arguments passed to the ProfileReport method used
             to create the EDA report. Ignored if ``profile`` is False.
 
+        
+        upper_clamp: int or float, default = None
+            Upper limit (inclusive) of values predicted/forecasted by models.
+
+
+        lower_clamp: int or float:, default = None
+            Lower limit (inclusive) of values predicted/forecasted by models.
+
 
         Returns:
             Global variables that can be changed using the ``set_config`` function.
@@ -1134,6 +1144,10 @@ class TimeSeriesExperiment(_SupervisedExperiment):
         self.strictly_positive = np.all(data_[target_name] > 0)
 
         self.enforce_pi = enforce_pi
+
+        # set clamp values
+        self.upper_clamp: Optional[Union[float,int]] = upper_clamp
+        self.lower_clamp: Optional[Union[float,int]] = lower_clamp
 
         return super().setup(
             data=data_,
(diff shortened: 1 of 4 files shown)


AJarman:
I have found the source of the failing tests -  this occurs is when a model is saved, then loaded after being saved, but loaded into into a fresh `TimeSeriesExperiment` Object which has not had `setup` called, therefore upper_clamp and lower_clamp values are never initialised (as they are only ever set or referenced outside the constructor).

I have a few potential solutions:

- It might be a more straightforward implementation to add these (optional) parameters to `TimeSeriesExperiment.predict_model()` and the functional API function.
    - Additionally this would mean that the Experiment attributed can be overridden/ignored if no longer relevant/correct without re-doing setup and any model training. 
    - We could then remove this from `setup()` altogether if desired.
-  We could set both upper and lower clamp values to `None` in the `TimeSeriesExperiment` constructor, this would then remove the `AttributeError` when calling the API.

AJarman:
Note that at the moment the tests are passing, however the implementation is checking is setup() has been called on the current experiment before applying the clamps.

Happy to change/enhance the implementations to those suggested above.

Yard1:
@AJarman Could we have the clamps be applied through a sktime wrapper/meta estimator, similarly to what was done for classification probability threshold? https://github.com/pycaret/pycaret/pull/1858

AJarman:
> @AJarman Could we have the clamps be applied through a sktime wrapper/meta estimator, similarly to what was done for classification probability threshold? That way the information about clamps would both be persisted with the model and also applied outside of pycaret, if needed. see #1858

I understand the need to be persisted across experiments/models, worth discussing

is there a time series equivalent to the superclasses defined here?:
`class CustomProbabilityThresholdClassifier(ClassifierMixin, BaseEstimator):`

Yard1:
@ngupta23 Any idea?

ngupta23:
Hi All, Sorry, I did not think about the persistence aspect. Not sure if this is avaiable in `sktime`. @ltsaprounis Do you have any idea if this could be implemented? Thanks!

AJarman:
@ngupta23 
I'm quite happy to refactor this to only use the clamp kwargs in the .predict() method and remove from setup(). 

This seems like the easiest first iteration to deliver the functionality/value without the persistence as I'm not sure why someone would do that. 

We can keep the issue/open a related issue for someone to implement the wrapper/metaEstimator?

ngupta23:
> I'm quite happy to refactor this to only use the clamp kwargs in the .predict() method and remove from setup().

The problem is that it is not only the user that needs to control the upper and lower limit in the predict_model method, but that we use predict_model internally as well in various places. So if we are to limit this, we need to add this to the setup function itself so it can be used everywhere internally (including calculation of metrics). In addition, as @Yard1 pointed out, this needs to be persisted as well in the model. So just adding it as an argument to predict_model method would not be sufficient.

We may have to table this feature unless there is a way to add it as a `sktime` model wrapper itself.

ngupta23:
@AJarman - I spoke to @ltsaprounis yesterday. He is going to check if the clamps can be added as `sktime` transformers (a new feature in `sktime`). We should table this development until we hear back about it.

ngupta23:
@AJarman , could you provide me a link to your LinkedIn Account so we can discuss future development efforts? Also, if you can join the pycaret slack, it would be great to discuss and align. Thanks!

https://join.slack.com/t/pycaret/shared_invite/zt-row9phbm-BoJdEVPYnGf7_NxNBP307w

PR #2041, opened 2022-01-09
revamped optimize_threshold function in classification module

moezali1:
Completely revamped `optimize_threshold` function. Only impact `pycaret.classification` module.

Prior to this revamp, the `optimize_threshold` function uses metrics on the hold-out set to determine the best metric. 

It also required `true_positive`, `false_positive`, `true_negative`, and `false_negative` to be entered as a parameter to the `optimize_threshold` function.

This revamp makes use of the `probability_threshold` parameter in the `create_model` (added in 2.3.5) and also uses the metrics from the metric container as opposed to asking the user to define tp,fp,tn,fn. It also works with custom metrics added to pycaret through the `add_metric` function.

Finally, previously we were only returning the float (Example: 0.66) at which point the custom cost function is optimized. Now we are returning the final object which is a PyCaret's custom class.

image

The only thing pending is the display monitor, as the loop runs it does not show anything and sometimes on the bigger dataset, this could be a while, especially if `grid_interval` is a very low number. 

@Yard1 See if you can quickly add a display monitor. I couldn't figure out the custom class thing. Sorry.

### DIFF
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -1883,7 +1883,7 @@ def optimize_threshold(
     optimize: str = "Accuracy",
     grid_interval: float = 0.1,
     return_data: bool = False, 
-    plot_kwargs: dict = {},
+    plot_kwargs: Optional[dict] = None,
 ):
 
     """
--- a/pycaret/classification.py
+++ b/pycaret/classification.py
@@ -1880,71 +1880,67 @@ def calibrate_model(
 
 def optimize_threshold(
     estimator,
-    true_positive: int = 0,
-    true_negative: int = 0,
-    false_positive: int = 0,
-    false_negative: int = 0,
-    grid_interval: float = 0.0001,
+    optimize: str = "Accuracy",
+    grid_interval: float = 0.1,
+    return_data: bool = False, 
+    plot_kwargs: dict = {},
 ):
 
     """
-    This function optimizes probability threshold for a given estimator using
-    custom cost function. The function displays a plot of optimized cost as a
-    function of probability threshold between 0.0 to 1.0 and returns the
-    optimized threshold value as a numpy float.
+    This function optimizes probability threshold for a trained classifier. It 
+    iterates over performance metrics at different ``probability_threshold`` with
+    a step size defined in ``grid_interval`` parameter. This function will display
+    a plot of the performance metrics at each probability threshold and returns the
+    best model based on the metric defined under ``optimize`` parameter.
 
 
     Example
     -------
     >>> from pycaret.datasets import get_data
     >>> juice = get_data('juice')
-    >>> from pycaret.classification import *
-    >>> exp_name = setup(data = juice,  target = 'Purchase')
+    >>> experiment_name = setup(data = juice,  target = 'Purchase')
     >>> lr = create_model('lr')
-    >>> optimize_threshold(lr, true_negative = 10, false_negative = -100)
+    >>> best_lr_threshold = optimize_threshold(lr)
 
 
-    estimator: scikit-learn compatible object
-        Trained model object
+    Parameters
+    ----------
+    estimator : object
+        A trained model object should be passed as an estimator.
 
 
-    true_positive: int, default = 0
-        Cost function or returns for true positive.
+    optimize : str, default = 'Accuracy'
+        Metric to be used for selecting best model. 
 
 
-    true_negative: int, default = 0
-        Cost function or returns for true negative.
+    grid_interval : float, default = 0.0001
+        Grid interval for threshold grid search. Default 10 iterations.
 
 
-    false_positive: int, default = 0
-        Cost function or returns for false positive.
+    return_data :  bool, default = False
+        When set to True, data used for visualization is also returned.
 
 
-    false_negative: int, default = 0
-        Cost function or returns for false negative.
+    plot_kwargs :  dict, default = {} (empty dict)
+        Dictionary of arguments passed to the visualizer class.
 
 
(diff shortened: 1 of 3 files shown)


pycaret/internal/tabular.py
```diff
@@ -8391,139 +8389,71 @@ def optimize_threshold(
...
+        results_df.append(model_results)
+
+    logger.info("optimization loop finished successfully")
+
+    results_concat = pd.concat(results_df, axis=0)
+    results_concat_melted = results_concat.melt(id_vars = ['probability_threshold'], value_vars=list(results_concat.columns[:-1]))
+    optimized_metric_index = np.array(results_concat_melted[results_concat_melted['variable'] == optimize]['value']).argmax()
+    best_model_by_metric = models_by_threshold[optimized_metric_index]
+
+    logger.info("plotting optimizartion threshold using plotly")
```
Yard1, line 8439:
```suggestion
    logger.info("plotting optimization threshold using plotly")
```

pycaret/internal/tabular.py
```diff
@@ -8391,139 +8389,71 @@ def optimize_threshold(
...
-        )
-        logger.info("Figure ready for render")
-        fig.show()
-    print(f"Optimized Probability Threshold: {t} | Optimized Cost Function: {y1}")
-    logger.info(
-        "optimize_threshold() succesfully completed......................................"
-    )
-
-    return float(t)
+        model = create_model_supervised(estimator, verbose=False, system=False, probability_threshold=i)
```
Yard1, line 8423: this is pretty heavy. in order to optimize prediction threshold we should only need to fit once (per fold). that being said it should be refactored in a future PR

pycaret/internal/tabular.py
```diff
@@ -8391,139 +8389,71 @@ def optimize_threshold(
...
+    results_concat = pd.concat(results_df, axis=0)
+    results_concat_melted = results_concat.melt(id_vars = ['probability_threshold'], value_vars=list(results_concat.columns[:-1]))
+    optimized_metric_index = np.array(results_concat_melted[results_concat_melted['variable'] == optimize]['value']).argmax()
+    best_model_by_metric = models_by_threshold[optimized_metric_index]
+
+    logger.info("plotting optimizartion threshold using plotly")
+
+    # plotting threshold
+    import plotly.express as px
+    title = str(model_name) + " Probability Threshold Optimization (default = 0.5) "
```
Yard1, line 8443:
```suggestion
    title = f"{model_name} Probability Threshold Optimization (default = 0.5)"
```

pycaret/internal/tabular.py
```diff
@@ -8391,139 +8389,71 @@ def optimize_threshold(
...
-
-    predicted = model.predict_proba(X_test)
-    predicted = predicted[:, 1]
+    if isinstance(optimize, str):
+        # checking optimize parameter
+        optimize = _get_metric(optimize).display_name
+        if optimize is None:
+            raise ValueError(
+                "Optimize method not supported. See docstring for list of available parameters."
+            )
```
Yard1, line 8403:
```suggestion
        optimize = _get_metric(optimize)
        if optimize is None:
            raise ValueError(
                "Optimize method not supported. See docstring for list of available parameters."
            )
        optimize = optimize.display_name
```

pycaret/classification.py
```diff
@@ -1880,71 +1880,67 @@ def calibrate_model(
...
     estimator,
-    true_positive: int = 0,
-    true_negative: int = 0,
-    false_positive: int = 0,
-    false_negative: int = 0,
-    grid_interval: float = 0.0001,
+    optimize: str = "Accuracy",
+    grid_interval: float = 0.1,
+    return_data: bool = False, 
+    plot_kwargs: dict = {},
```
Yard1, line 1886:
```suggestion
    plot_kwargs: Optional[dict] = None,
```

pycaret/internal/tabular.py
```diff
@@ -8302,61 +8302,59 @@ def calibrate_model(
...
     estimator,
-    true_positive: int = 0,
-    true_negative: int = 0,
-    false_positive: int = 0,
-    false_negative: int = 0,
-    grid_interval: float = 0.0001,
+    optimize: str = "Accuracy",
+    grid_interval: float = 0.1,
+    return_data: bool = False, 
+    plot_kwargs: dict = {},
```
Yard1, line 8308:
```suggestion
    plot_kwargs: Optional[dict] = None,
```

pycaret/internal/tabular.py
```diff
@@ -8391,139 +8389,71 @@ def optimize_threshold(
...
+    optimized_metric_index = np.array(results_concat_melted[results_concat_melted['variable'] == optimize]['value']).argmax()
+    best_model_by_metric = models_by_threshold[optimized_metric_index]
+
+    logger.info("plotting optimizartion threshold using plotly")
+
+    # plotting threshold
+    import plotly.express as px
+    title = str(model_name) + " Probability Threshold Optimization (default = 0.5) "
+    fig = px.line(results_concat_melted, x="probability_threshold", y="value", title = title,\
+                    color='variable', **plot_kwargs)
```
Yard1, line 8445:
```suggestion
    plot_kwargs = plot_kwargs or {}
    fig = px.line(results_concat_melted, x="probability_threshold", y="value", title = title,\
                    color='variable', **plot_kwargs)
```

Yard1 (approved):

pycaret/internal/tabular.py
moezali1: Ok.

Issue #2054, opened 2022-01-12
[BUG] error on check_fairness


reza1615:
It seems line 10422 in tabular.py the .iloc should be .loc
![image](https://user-images.githubusercontent.com/2202483/149231650-f0c0fb17-4b7d-40ec-b906-b0f45176dd9b.png)

moezali1:
@reza1615 Thank you. The PR has been merged now.

PR #2055, opened 2022-01-13
resolving bug when index is not ordinal number

reza1615:
Resolving bug #2054

```
income = get_data('income')
```
```
import random
income.index=random.sample(range(200000, 300000), income.shape[0])
```
```
exp_name = setup(data = income,  target = 'income >50K')
lightgbm = create_model('lightgbm')
lightgbm_fairness = check_fairness(lightgbm, sensitive_features = ['sex', 'race'])
```

image


moezali1:
@reza1615 Thanks. Approved.

Issue #2053, opened 2022-01-12
Bug in dahsboard () when the columns has especial characters


reza1615:
It seems before sending feature columns to the explainerdashboard. pycaret should clean them

```
InvalidComponentIdError: The element `feature-input-CE_Send to Fraud_1.0-input-40` contains `.` in its ID.
Characters `.`, `{` are not allowed in IDs
```

![image](https://user-images.githubusercontent.com/2202483/149231429-13b53e3d-4035-4419-a912-40c5e67b1324.png)

moezali1:
PR merged.

PR #2058, opened 2022-01-14
Replace unsupported characters in df

reza1615:
fixing #2053


moezali1:
@reza1615 There are two functions one for regression `_create_regression_dashboard`. You will have to do the same change there as well!

reza1615:
@moezali1 Done

moezali1 (approved):

PR #2097, opened 2022-01-26
Bugfix for displaying inferred datatypes

swaldow:
This is my first contribution to a git project, I hope it's all ok.
## Related Issuse or bug

https://pycaret.slack.com/archives/C01HTQAAJER/p1642087232044700

Info about Issue or bug

If the dataset contains more than 60 columns, then the `setup()` function cannot display all inferred datatypes, because the intermediate output of the inferred datatypes is truncated.

Increasing the output with `pd.options.display.max_rows` does not work because the maximum number of rows is reset within the execution of `setup()`.

I have found the place in the code of `preprocess.py` where `display(dt_print_out[["Data Type"]])`
is executed and the inferred datatypes will be given out. I added 
`pd.set_option("display.max_rows", 1000)` before the execution of the `display(dt_print_out[["Data Type"]])` function and 
`pd.reset_option("display.max_rows")` after the excecution.

## Type of change

## How Has This Been Tested?

Passes all existing unittests.

NA

## Checklist:

- [x] ~~New~~ and existing unit tests pass locally with my changes.

I don't think the missing items on the checklist need to be ticked off, but if there is anything that needs to be done let me know.

## Screenshots

 Original           | Updated
Before | After

### DIFF
--- a/pycaret/internal/preprocess.py
+++ b/pycaret/internal/preprocess.py
@@ -319,7 +319,14 @@ class DataTypes_Auto_infer(BaseEstimator, TransformerMixin):
             # if we added the dummy  target column , then drop it
             dt_print_out.drop(index="dummy_target", errors="ignore", inplace=True)
 
+            # increase maximum displayed rows to 1000
+            pd.set_option("display.max_rows", 1000)
+
             display(dt_print_out[["Data Type"]])
+
+            # reset pandas option
+            pd.reset_option("display.max_rows")
+            
             self.response = input()
 
             if self.response in [


ngupta23:
I am not sure this should be hard coded in the code. Not everyone would want to see so many rows. On the other hand, some users may want to see more than 1000. Should this be an argument in `setup` itself?

```python
pd.set_option("display.max_rows", 1000)
```

moezali1:
@ngupta23 If 1000 is a small number I would rather set it to 99999 or something like that instead of parameterizing it since there isn't much value of this parameter. The only reason we want to increase it is that in pandas the default number of rows displayed is 60 so when you run the setup, the confirmation grid is truncated because of `max_display_rows`. This wouldn't be a problem in 3.0 since we are completely removing the display grid.

tvdboom:
This will all change in v3.0, where we don't show the data type per feature anymore but rather the number of features per data type. I agree with @ngupta23 that setting it to 1000 is not a sustainable solution. I would leave it as is and wait for v3.0.

moezali1:
Based on @tvdboom feedback, I am closing this PR without merging.

PR #2101, opened 2022-01-29
Time series models to use sp_to_use

### LINKED ISSUE #1894: Time Series models to use `sp_to_use`
Currently, all models use something like this

```python
self.seasonality_present = globals_dict.get("seasonality_present")
sp = globals_dict.get("seasonal_period")
self.sp = sp if sp is not None else 1
```

```python
def _set_args(self) -> Dict[str, Any]:
    args = {"sp": self.sp} if self.seasonality_present else {}
```

Consider replacing `sp` here with `sp_to_use` instead (from `self.sp_to_use`).

```python
self.sp_to_use = globals_dict.get("sp_to_use")

def _set_args(self) -> Dict[str, Any]:
    args = {"sp": self.sp_to_use}
```

drmario-gh:
## Related Issuse or bug

Closes #1894 

For all time series models that handle seasonality, I make them use `sp_to_use` instead of the `seasonal_period` specified in the experiment setup. 

With this change a couple of None checks were irrelevant, so I removed them as well (indicated with comments on the PR). 

## Type of change

## How Has This Been Tested?
Unit tests (no new unit tests were added). 

`NA`

## Checklist:
Not sure about this one: black pre-commit hook introduced changes in parts of the code I did not modify (in files I modify) and I ended up removing them to avoid mixing them with my changes. Should I reintroduce them in this PR?

### DIFF
--- a/pycaret/internal/plots/time_series.py
+++ b/pycaret/internal/plots/time_series.py
@@ -812,18 +812,12 @@ def plot_time_series_decomposition(
     decomp_result = None
     data_ = data.to_timestamp() if isinstance(data.index, pd.PeriodIndex) else data
 
-    sp_to_use = data_kwargs.get("sp_to_use", None)
+    sp_to_use = data_kwargs.get("sp_to_use")
     if plot == "decomp_classical":
-        if sp_to_use is None:
-            decomp_result = seasonal_decompose(data_, model=classical_decomp_type)
-        else:
             decomp_result = seasonal_decompose(
                 data_, period=sp_to_use, model=classical_decomp_type
             )
     elif plot == "decomp_stl":
-        if sp_to_use is None:
-            decomp_result = STL(data_).fit()
-        else:
             decomp_result = STL(data_, period=sp_to_use).fit()
 
     fig = make_subplots(
--- a/pycaret/tests/conftest.py
+++ b/pycaret/tests/conftest.py
@@ -56,6 +56,7 @@ def load_ts_models(load_setup):
         "gpu_param": False,
         "X_train": pd.DataFrame(get_data("airline")),
         "enforce_pi": False,
+        "sp_to_use": 1,
     }
     ts_models = get_all_model_containers(globals_dict)
     ts_experiment = load_setup
--- a/pycaret/tests/conftest.py
+++ b/pycaret/tests/conftest.py
@@ -56,7 +56,7 @@ def load_ts_models(load_setup):
         "gpu_param": False,
         "X_train": pd.DataFrame(get_data("airline")),
         "enforce_pi": False,
-        "sp_to_use": 1,
+        "sp_to_use": 12,
     }
     ts_models = get_all_model_containers(globals_dict)
     ts_experiment = load_setup
--- a/pycaret/tests/time_series_test_utils.py
+++ b/pycaret/tests/time_series_test_utils.py
@@ -207,10 +207,10 @@ def _return_data_with_without_period_index():
 
 def _return_model_names_for_plots():
     """Returns models to be used for testing plots. Needs
-    - 1 model that has prediction interval ("theta")
-    - 1 model that does not have prediction interval ("lr_cds_dt")
-    - 1 model that has in-sample forecasts ("theta")
-    - 1 model that does not have in-sample forecasts ("lr_cds_dt")
+        - 1 model that has prediction interval ("theta")
+        - 1 model that does not have prediction interval ("lr_cds_dt")
+        - 1 model that has in-sample forecasts ("theta")
+        - 1 model that does not have in-sample forecasts ("lr_cds_dt")
     """
     model_names = ["theta", "lr_cds_dt"]
     return model_names
--- a/pycaret/tests/time_series_test_utils.py
+++ b/pycaret/tests/time_series_test_utils.py
@@ -87,7 +87,7 @@ def _get_seasonal_values_alphanumeric():
     choice_list = ['10','20','30','40','50','60']
     #prefix = random.choice(choice_list)
     return [(random.choice(choice_list),k,v.value) for k, v in SeasonalPeriod.__members__.items()]
- 
+
 def _check_windows():
     """Check if the system is Windows."""
     import sys
@@ -107,6 +107,7 @@ def _return_model_names():
         "X_train": pd.DataFrame(get_data("airline")),
         "enforce_pi": False,
         "seasonal_period": 2,
+        "sp_to_use": 2,
     }
     model_containers = get_all_model_containers(globals_dict)
 
@@ -206,10 +207,10 @@ def _return_data_with_without_period_index():
 
 def _return_model_names_for_plots():
     """Returns models to be used for testing plots. Needs
-        - 1 model that has prediction interval ("theta")
-        - 1 model that does not have prediction interval ("lr_cds_dt")
-        - 1 model that has in-sample forecasts ("theta")
-        - 1 model that does not have in-sample forecasts ("lr_cds_dt")
+    - 1 model that has prediction interval ("theta")
+    - 1 model that does not have prediction interval ("lr_cds_dt")
+    - 1 model that has in-sample forecasts ("theta")
+    - 1 model that does not have in-sample forecasts ("lr_cds_dt")
     """
     model_names = ["theta", "lr_cds_dt"]
     return model_names
--- a/pycaret/containers/models/time_series.py
+++ b/pycaret/containers/models/time_series.py
@@ -856,12 +856,14 @@ class ExponentialSmoothingContainer(TimeSeriesContainer):
             }
         return tune_distributions
 
+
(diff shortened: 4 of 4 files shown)


pycaret/tests/conftest.py
```diff
@@ -56,6 +56,7 @@ def load_ts_models(load_setup):
         "gpu_param": False,
         "X_train": pd.DataFrame(get_data("airline")),
         "enforce_pi": False,
+        "sp_to_use": 1,
```
drmario-gh, line 59: By definition, `sp_to_use` will at least be 1, so I add it in this fixture.

pycaret/tests/time_series_test_utils.py
```diff
@@ -107,6 +107,7 @@ def _return_model_names():
         "X_train": pd.DataFrame(get_data("airline")),
         "enforce_pi": False,
         "seasonal_period": 2,
+        "sp_to_use": 2,
```
drmario-gh, line 110: And same here, by definition, `sp_to_use` will at least be 1, so I add it in this fixture.

pycaret/tests/conftest.py
ngupta23: Will it not be 12 in this case? Since data is monthly seasonal (airline dataset)?

pycaret/internal/plots/time_series.py
```diff
@@ -812,18 +812,12 @@ def plot_time_series_decomposition(
     decomp_result = None
     data_ = data.to_timestamp() if isinstance(data.index, pd.PeriodIndex) else data
 
-    sp_to_use = data_kwargs.get("sp_to_use", None)
+    sp_to_use = data_kwargs.get("sp_to_use")
     if plot == "decomp_classical":
-        if sp_to_use is None:
-            decomp_result = seasonal_decompose(data_, model=classical_decomp_type)
-        else:
             decomp_result = seasonal_decompose(
```
ngupta23, line 817: Could you please confirm that this works when `sp_to_use` = 1. If so, then this looks good. Thanks!

pycaret/internal/plots/time_series.py
```diff
@@ -812,18 +812,12 @@ def plot_time_series_decomposition(
...
                 data_, period=sp_to_use, model=classical_decomp_type
             )
     elif plot == "decomp_stl":
-        if sp_to_use is None:
-            decomp_result = STL(data_).fit()
-        else:
             decomp_result = STL(data_, period=sp_to_use).fit()
```
ngupta23, line 821: Could you please confirm that this works when `sp_to_use = 1`. If so, then this looks good. Thanks!

ngupta23:
Looks good. Just left some minor comments to check. Will merge after you confirm and all tests pass.

pycaret/tests/conftest.py
drmario-gh: 12 definitely makes sense, because for the `airline` dataset, if we don't specify a `seasonal_period`, that's the seasonality that would be inferred from the data and the value that `sp_to_use` will end up having. But that would change the tests as they were before my change and I didn't wanted to introduce such change silently. 

Before my change, all models had:
https://github.com/pycaret/pycaret/blob/e153069683166a2096567e5e4b01df4132e4ad65/pycaret/containers/models/time_series.py#L513-L514
which means that `self.sp` ended up being 1 in the tests.

Other than that, I see no problem in changing the value. Let me know what you think.

pycaret/internal/plots/time_series.py
drmario-gh: The thing here is that `sp_to_use` cannot be None. So L818 was unreachable code really.

`seasonal_decompose` does work when `sp_to_use`, although the result is useless: a classical decomposition with no seasonality is all trend. Here is the classical decomposition for the airline data, if we impose `seasonal_period=1` and thus `sp_to_use=1``: 
classical_sp_1

pycaret/internal/plots/time_series.py
drmario-gh: As in the classical decomposition, since `sp_to_use` cannot be None, L825 was unreachable code really. But in this case, the STL decomposition does not work with `sp_to_use = 1`: statsmodels code does not allow it (and it makes sense, because it would be pointless otherwise):
```
statsmodels/tsa/_stl.pyx in statsmodels.tsa._stl.STL.__init__()
ValueError: period must be a positive integer >= 2
```
Note also that this is already happening, it is no new behavior. I'd leave this behavior as is: it makes sense for this to fail if `sp_to_use = 1` and the error is pretty informative.

drmario-gh:
> Looks good. Just left some minor comments to check. Will merge after you confirm and all tests pass.

Great. Answered your questions and I just have a small doubt on your first comment.

pycaret/tests/conftest.py
ngupta23: But `sp_to_use` would have been 12 before the change as well (not 1). So before the change, `self.sp` (in the model container) would end up taking a value of 12 also (not 1). Did I miss something here?

image

pycaret/internal/plots/time_series.py
ngupta23: Fair enough. This is a bug though and we will need to fix this. The appropriate action would be to through a warning message in this case from pycaret rather than have the program crash. I will open a new issue for this.

Thanks for catching this!

pycaret/tests/time_series_test_utils.py
ngupta23: Did you mean `sp_to_use` will be 2 and not 1?

pycaret/tests/conftest.py
ngupta23: You can confirm this by creating an ARIMA model (using `time_series` branch which does not have your changes yet). You will see that it will take sp = 12 in the model.

drmario-gh: Yes, this one is a bit tricky. I think we are both right but talking about different things.
* Yes, following the regular execution path, if you don't specify `seasonal_period` in the setup, `sp_to_use` is inferred from the data, and ends up being 12.
* However, if we take a look at one of the tests that fail if I don't add this `sp_to_use=1`:
https://github.com/pycaret/pycaret/blob/b1502037b87d035b5d75585a31eb1ac0e9cdce7e/pycaret/tests/test_time_series_blending.py#L43
it uses the fixture `load_models`, which in turn calls `get_all_model_containers`, triggering the init method of all time series models, and it does so with a `globals_dict` that cannot really happen in a regular code execution path: it has no `sp_to_use`.  
https://github.com/pycaret/pycaret/blob/b1502037b87d035b5d75585a31eb1ac0e9cdce7e/pycaret/tests/conftest.py#L78-L86
Thus, before my changes, `self.sp` ends up being 1 in the models initialization:
https://github.com/pycaret/pycaret/blob/b1502037b87d035b5d75585a31eb1ac0e9cdce7e/pycaret/containers/models/time_series.py#L399-L400
After my changes, however, models initialization don't accept a `self.sp=None` and fail if I don't specify an `sp_to_use` in `load_models``:
https://github.com/pycaret/pycaret/blob/e153069683166a2096567e5e4b01df4132e4ad65/pycaret/containers/models/time_series.py#L399
https://github.com/pycaret/pycaret/blob/e153069683166a2096567e5e4b01df4132e4ad65/pycaret/containers/models/time_series.py#L434

This is inconsequential really because in the `load_models` fixture, `load_setup` is called afterwards, so the regular setup method is called and models are finally created through `create_model`, having the right `sp_to_use` which will be 12, no matter the value specified in `load_models`.

https://github.com/pycaret/pycaret/blob/b1502037b87d035b5d75585a31eb1ac0e9cdce7e/pycaret/tests/conftest.py#L65-L72

So tl;dr, it really does not matter the value that we set up here, it is only used in the models init triggered by `get_all_model_containers` , and the call `to `get_all_model_containers` is only used to get all model ids. These ids are then used to create proper model instances which will have the right `sp_to_use=12` value. But what will make more sense probably is directly specifying here `sp_to_use=12`. I'm changing the value to 12.

pycaret/tests/time_series_test_utils.py
drmario-gh: Yes, sorry, what I wanted to say really is that `sp_to_use` depends on whether the seasonality check for `seasonal_period=2` is positive (in which case `sp_to_use=2`) or negative (in which case `sp_to_use=1`).

In this case, we know that such check is positive and then `sp_to_use=2`.

ngupta23: ok, thanks!

pycaret/tests/conftest.py
ngupta23: Thank you for the detailed analysis, clarifying, and changing it to 12. I think at a high level it sounds reasonable, but the only issue might be that when `sp_to_use = 1`, some models like `snaive` (seasonal naive) are disabled. But `sp_to_use = 12` (value > 1) might be safer for that reason.

PR #2112, opened 2022-02-01
Solving installing UTF error

reza1615:
resolving the below error
image


moezali1:
@reza1615 Can you please only keep this PR for UTF error and do a separate PR for drift report, please. Thanks

PR #2237, opened 2022-03-03
Return train score

### LINKED ISSUE #1909: Requesting for training scores along with test scores after training models, tuning models or comparing models etc
1. Currently, after running the functions create_model, compare_model function, etc., do not return train scores, only test scores across the splits or the mean of the test scores are displayed. 
2. What if someone wants to validate the scores coming in the training sets, or how the scores are varying across the training sets? The comparison between training scores and testing scores helps us understand how the model is performing in terms of fitting, for example, overfitting, underfitting, or covariate shift. This is very crucial in the case of Time aware cross-validation.  
3. I have introduced a parameter namely return_train_score, which is a boolean variable and helps the user to specify if they want the training score to return or not. So, the choice is optional and the training score is merged with the test data frame if the parameter is set as true. I have also updated the display formats accordingly and the scores are shown as Train_Score, Test_Score, etc. 

image    

4. After the modification, the function has the outputs in the following manner, the number of scores can easily be reduced with Pycaret's own remove_metric function.
    
image
    
image
    
5. If you find this useful, I can dedicate my time to work on this issue and any suggestions on this issue are most welcome.

### LINKED ISSUE #2236: [BUG] Help with generating metrics on Train data set
**Describe the bug**

**To Reproduce**

```python
##Setting up the environment in PyCaret
regr = setup(train, target = 'y',n_jobs=-1,test_data=test,numeric_features = features)

# compare all baseline models and select top 5
top5 = compare_models(n_select = 5)

# tune models
tuned_top5 = [tune_model(i, n_iter = 50) for i in top5]

# ensemble models
bagged_top5 = [ensemble_model(i) for i in tuned_top5]

# blend models
blender = blend_models(estimator_list = top5)

# stack models
stacker = stack_models(estimator_list = top5)

# select best model
best = automl(optimize = 'R2')
cv_metrics = pull()
print("CV metrics:")
print(cv_metrics)

# Deploy Model to generate predictions on train data
predict_model(best, train)
# pull
train_metrics = pull()
print("Train metrics:")
print(train_metrics)

# Deploy Model to generate predictions on hold out data
predict_model(best)
# pull
test_metrics = pull()
print("Test metrics:")
print(test_metrics)
```

**Expected behavior**

**Additional context**

**Versions**
<details>

</details>

levelalphaone:
## Incorporating return train score in supervised learning models

Closes #1909
Closes #2236 

I have modified the cross-validation scores dataframe, which is returned after create_model_supervised is called, as the specified format discussed by the pycaret team. Due to the changes I made, now the model results would show in a specific format when the pull function be called, namely CV-Train to show the score in the training split and CV-Val to show the scores in the Validation split. There is also a split name Train, which shows the final model results when the model is retained on the whole dataset, after cross-validation. I have also added return_train_score, a bool parameter to give the user the option whether to return the training score for the training sample in the cross-validation, or not.

## Type of change

## How Has This Been Tested?

I have run the pytest on the tests folder and checked if the test is passed or not in my current conda environment, with python 3.8.12
Furthermore, I have tested the code for two separate datasets, one from regression and another from classification. All the functions that have been changed in the pycaret-> regression.py or pycaret->classification.py have been tested using the example given there, with and without the added parameter.
You can run the same thing with setting the parameter return_train_score=True or False in the following functions - create_model, stack_model, tune_model, ensemble_model, blend_models, calibrate_model, finalize_model and, automl

Two test files, namely test_classification_parallel.py and test_regression_parallel.py are failing, where the test function itself is not called, but failing due to 'dask' parameter was passed in FugueBackend, which raised  TypeError: dask is not a type. I think something is off with the test file or the class FugueBackend itself. I am requesting you to please check the function.

## Checklist:

## Screenshots

 <b>original screenshot </b>

image

<b>updated screenshot </b>

image

image

image


pycaret/classification.py
```diff
@@ -927,6 +928,11 @@ def create_model(
         Score grid is not printed when verbose is set to False.
 
 
+    return_train_score: bool, default = False
+        If not False, will evaluate the train value scores.
```
ngupta23, line 932: I would simplify this a bit.

```
If False, returns the CV Validation scores only. 
If True, returns the CV training scores along with the CV validation scores.
This is useful when the user wants to do bias-variance tradeoff. A high CV
training score with a low corresponding CV validation score indicates overfitting.
```

pycaret/classification.py
```diff
@@ -927,6 +928,11 @@ def create_model(
...
+        Intended to be fed as an input from the user.
```
ngupta23, line 933: I would remove this line. Everything is intended to be fed by the user :)

pycaret/classification.py
```diff
@@ -1122,6 +1130,11 @@ def tune_model(
         print more messages. Ignored when ``verbose`` param is False.
 
 
+    return_train_score: bool, default = False
+        If not False, will evaluate the train value scores.
```
ngupta23, line 1134: Same change as above.

pycaret/classification.py
```diff
@@ -1247,6 +1262,11 @@ def ensemble_model(
         Score grid is not printed when verbose is set to False.
 
 
+    return_train_score: bool, default = False
+        If not False, will evaluate the train value scores.
+        Intended to be fed as an input from the user.
```
ngupta23, line 1267: Same changes as above

pycaret/classification.py
```diff
@@ -1364,6 +1386,11 @@ def blend_models(
...
```
ngupta23, line 1391: Same changes as above

pycaret/classification.py
```diff
@@ -1897,6 +1933,11 @@ def calibrate_model(
...
+
```
ngupta23, line 1939: Same changes as above

pycaret/classification.py
```diff
@@ -2125,6 +2168,10 @@ def finalize_model(
         Dictionary of tag_name: String -> value: (String, but will be string-ified if
         not) passed to the mlflow.set_tags to add new custom tags for the experiment.
 
+    return_train_score: bool, default = False
+        If not False, will evaluate the train value scores.
+        Intended to be fed as an input from the user.
```
ngupta23, line 2173: Same changes as above

pycaret/classification.py
```diff
@@ -2371,11 +2419,16 @@ def automl(optimize: str = "Accuracy", use_holdout: bool = False) -> Any:
         When set to True, metrics are evaluated on holdout set instead of CV.
 
 
+    return_train_score: bool, default = False
+        If not False, will evaluate the train value scores.
+        Intended to be fed as an input from the user.
```
ngupta23, line 2424: Same changes as above

pycaret/internal/tabular.py
```diff
@@ -4,6 +4,7 @@
 # Release: PyCaret 2.2
 # Last modified : 26/08/2020
 
+from cgi import test
```
ngupta23, line 7: Is this import needed?

pycaret/internal/tabular.py
```diff
@@ -2922,6 +2924,10 @@ def create_model_supervised(
         If not None, will use this dataframe as training target.
         Intended to be only changed by internal functions.
 
+    return_train_score: bool, default = False
+        If not False, will evaluate the train value scores.
+        Intended to be fed as an input from the user.
```
ngupta23, line 2929: Same changes as above in all places where this is documented.

moezali1:
Hi, I have a question. Why the output for both this code is different even though the code is the same. Sorry I am lost.

image

pycaret/internal/tabular.py
```diff
@@ -3257,21 +3287,68 @@ def create_model_supervised(
             logger.info("Finalizing model")
             with io.capture_output():
                 pipeline_with_model.fit(data_X, data_y, **fit_kwargs)
+            
+            if isinstance(model, CustomProbabilityThresholdClassifier):
```
ngupta23, line 3291: Could you add some notes here as to what is happening here? It will help to review this.

ngupta23:
@levelalphaone - so far this looks good to me. Just some minor comments about the docstring (which needs to be replaced in multiple places).

I did not understand one change. Maybe you can add some code comments there to elaborate on what you are doing.

@Yard1 I have reviewed from my side. Can you also do a quick check? Thanks!

moezali1:
@ngupta23 Would u know the answer to what I asked above?

moezali1:
@Yard1 This will go in the 2.3.7 release.

ngupta23:
> @ngupta23 Would u know the answer to what I asked above? 

Not sure what the entire code looks like. One is regression, one is classification.

pycaret/classification.py
levelalphaone: Sure, working on this.

pycaret/classification.py
levelalphaone: Yes, absolutely!

pycaret/internal/tabular.py
levelalphaone: I did not add this particular line. I simply fetched the latest push in the main repo and added my code, I have not added any import in any file.

levelalphaone:
> Hi, I have a question. Why the output for both this code is different even though the code is the same. Sorry I am lost.
> 
> image

Hi @moezali1. The code is run with different pycaret setup and dataframe. The first one is for classification and the second one is for regression. Could not show the entire code here, but you get the gist, right? If you need, I can share the notebooks as well.

pycaret/internal/tabular.py
levelalphaone: Sure @ngupta23. Thanks for the comments.

pycaret/internal/tabular.py
```diff
@@ -3217,37 +3227,57 @@ def create_model_supervised(
...
+                    "Fold": np.arange(fold).tolist() + ["Mean", "Std"],
+            })
 
-        # yellow the mean
-        model_results = color_df(model_results, "yellow", ["Mean"], axis=1)
-        model_results = model_results.set_precision(round)
+        model_scores = pd.concat([pd.DataFrame(score_dict), pd.DataFrame(avgs_dict)]).reset_index(drop=True)
+
+        model_results = pd.concat([model_results, model_scores], axis=1)
+        model_results.set_index(['Split', 'Fold'], inplace=True)
```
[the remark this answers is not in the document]
levelalphaone: The coloring is now done afterward like this - 

image

pycaret/internal/tabular.py
levelalphaone: Yes, I would add some notes here.

levelalphaone:
> @levelalphaone - so far this looks good to me. Just some minor comments about the docstring (which needs to be replaced in multiple places).
> 
> I did not understand one change. Maybe you can add some code comments there to elaborate on what you are doing.
> 
> @Yard1 I have reviewed from my side. Can you also do a quick check? Thanks!

Yes, sure. The doubt you had there is about calculating the score for the final model. That piece of code takes the model and calculates the score based on the whole training data and concatenates the information in the model_esults dataframe. I will add pieces of information in the code as well, to halp everyone understand.

pycaret/internal/tabular.py
```diff
@@ -3257,21 +3289,74 @@ def create_model_supervised(
             logger.info("Finalizing model")
             with io.capture_output():
                 pipeline_with_model.fit(data_X, data_y, **fit_kwargs)
+            
+            # calculating model results with final model on complete train dataset
```
ngupta23, line 3293: @Yard1  Could you please review this section? Can we simply get the metrics by doing a pull after the final fit or do we need to do all these steps?

ngupta23:
@levelalphaone Looks good from my side. I have @Yard1 to give some sections another look since he is more familiar with these modules. Once his review is done, we can merge this.

Thanks!

ngupta23 (approved):
Requested @Yard1  to review certain sections I am unfamiliar with. Rest looks ok.

pycaret/classification.py
```diff
@@ -2341,7 +2403,7 @@ def load_model(
     )
 
 
-def automl(optimize: str = "Accuracy", use_holdout: bool = False) -> Any:
+def automl(optimize: str = "Accuracy", use_holdout: bool = False, return_train_score: bool = False) -> Any:
```
Yard1, line 2406: has this been formatted with black?

pycaret/internal/tabular.py
Yard1: We should be able to use `create_model` here with set data arguments and then `pull`

Yard1 (changes requested):
This large block of code should be possible to replace with a call to `create_model`

pycaret/classification.py
levelalphaone: @Yard1 I have not formatted the code with black, will do it and then push again.

levelalphaone:
> This large block of code should be possible to replace with a call to `create_model`

@Yard1 I have only used the code from predict_model to have the score on the whole train dataset. I guess, what you are referring to is to the create_model with cross-validation set to false, and predict set to true. But predict_model only returns scores for the test set. So, even if you pass the train data, it does not evaluate on that, that is why I took the scorer portion from there and evaluated it on the train set. Please correct me if I am wrong. And please provide suggestions from your end as well.

pycaret/classification.py
levelalphaone: Should I format all files with black? Or this particular file only, because after I applied black in the root folder, 38 files were affected. Please let me know @Yard1

ngupta23: Only the files that you modified.

levelalphaone: Okay, sure!

levelalphaone: Done!

levelalphaone:
> > This large block of code should be possible to replace with a call to `create_model`
> 
> @Yard1 I have only used the code from predict_model to have the score on the whole train dataset. I guess, what you are referring to is to the create_model with cross-validation set to false, and predict set to true. But predict_model only returns scores for the test set. So, even if you pass the train data, it does not evaluate on that, that is why I took the scorer portion from there and evaluated it on the train set. Please correct me if I am wrong. And please provide suggestions from your end as well.

@Yard1 can you please look into this?

Yard1:
Hey @levelalphaone I have pushed a commit, please take a look

levelalphaone:
> Hey @levelalphaone I have pushed a commit, please take a look

@Yard1 I have understood the piece of code, that you have written and this way, it would work, and also it will help in passing any dataframe and then computing the score. I am modifying the code and I will create a commit.

levelalphaone:
@Yard1 I have looked into the code and saw that you have already made the necessary changes and the code works perfectly fine. Thanks for the support. I think this issue is now resolved. Thanks!

Issue #2137, opened 2022-02-07
Installation - Dependency cleanup


ngupta23:
**Is your feature request related to a problem? Please describe.**
Currently, there are many dependencies that can conflict with each other during installation.

**Describe the solution you'd like**
Break up installation into smaller modules

Ideas

- [x] Break into smaller modules
    - [x] pip install pycaret for Regression and Classification
    - [x] pip install pycaret[ts] for Time Series
    - [x] pip install pycaret[nlp] for Natural Language Processing
    - [x] pip install pycaret[anomaly] for Anomaly Detection
- [x] Make some dependencies optional for certain modules
    - [x] Yellowbricks is not needed for time series but can conflict with its installation  #2114  
    - [x] `check_fairness` library only applicable for regression/classification - should not be installed for other modules.

PR #2272, opened 2022-03-11
Refactor dependencies

tvdboom:
## Related Issuse or bug

Refactor dependencies. Part of issue #2137 
Possible to install subsets of dependencies.

## Type of change

## How Has This Been Tested?

## Checklist:

## Screenshots

 Original           | Updated


requirements.txt
```diff
@@ -1,29 +1,23 @@
...
-mlflow
-imbalanced-learn>=0.8.1
-category-encoders>=2.2.2
-scikit-plot #for lift and gain charts
-Boruta
\ No newline at end of file
+
+# Time-series
+statsmodels>=0.13.2
+sktime>=0.9.0
```
ngupta23, line 21: sktime>=0.10.1 per latest

ngupta23, line 21: sktime>=0.10.1 per latest

requirements.txt
```diff
@@ -1,29 +1,23 @@
...
-kmodes>=0.10.1
-mlflow
-imbalanced-learn>=0.8.1
-category-encoders>=2.2.2
-scikit-plot #for lift and gain charts
-Boruta
\ No newline at end of file
+
+# Time-series
+statsmodels>=0.13.2
```
ngupta23, line 20: statsmodels~=0.12.1 because of issues with earlier versions

ngupta23, line 20: statsmodels~=0.12.1 because of issues with earlier versions

requirements.txt
```diff
@@ -1,29 +1,23 @@
...
+sktime>=0.9.0
+tbats>=1.1.0
+pmdarima>=1.8.5
```
ngupta23, line 23: pmdarima>=1.8.0 will do

ngupta23, line 23: pmdarima>=1.8.0 will do

ngupta23, line 23: Also, add `kaleido`

requirements.txt
```diff
@@ -1,29 +1,23 @@
...
+imbalanced-learn>=0.8.1
+category-encoders==2.4.0
+lightgbm>=3.0.0
+
+# Plotting
+matplotlib>=3.3.0
+scikit-plot>=0.3.7
 yellowbrick>=1.4
-lightgbm>=2.3.1
 plotly>=5.0.0
```
ngupta23, line 17: Also, add `kaleido` after plotly

requirements.txt
tvdboom: what issues? All test pass on my pc with the newest version installed

ngupta23: Must have been resolved now. I would keep it same as what sktime has it as.

https://github.com/alan-turing-institute/sktime/blob/f25e4d906e723700351fff664771cb28dd244bc9/pyproject.toml#L51

setup_nightly.py
```diff
@@ -39,10 +39,22 @@ def readme():
...
-    extras_require={"full": optional_required,
-                    "test": test_required + optional_required},
+    extras_require={
+        "analysis": required_optional[1:9],
+        "models": required_optional[11:15],
+        "tuners": required_optional[17:22],
+        "mlops": required_optional[24:30],
+        "nlp": required_optional[32:38],
+        "full": required_optional,
+    },
```
Yard1, line 57: I think it would make more sense to split them into separate requirements files, or just define them directly in setup.py. This will break the moment we add or remove a dependency.

tvdboom: thought about it, but having 8 separate requirements files is unclear an unnecessary. Let's think if there's a way to keep the two files but make the code so that it doesn't break when changing requirements

Yard1: you could perhaps use a regex or some simple logic to grab the lines between two specially formatted comments?

tvdboom: I changed it. Now it should work as long as you keep the spaces between the sections

tvdboom: and keep the order of the sections. new dependencies can just be added beneath the respective section

ngupta23:
So, did we nuke `prophet`? If so, I think there are some others that were associated with it as well.

https://github.com/facebook/prophet/issues/1598

```python
LunarCalendar>=0.0.9
convertdate>=2.1.2
holidays>=0.13
```

ngupta23:
Actually, thinking about this a bit, I would actually prefer to have `prophet` in there as an optional dependency since the code has special handling for this library, and not having it in there would mean that this special handling would not get checked, so things like this would get missed out.

https://pipelines.actions.githubusercontent.com/MDDrRrtw7L4Hj0kTVUIzTl4Jqsb6FBoISWxstiMbSUrhYOUMiz/_apis/pipelines/1/runs/4047/signedlogcontent/10?urlExpires=2022-03-13T22%3A47%3A36.9610819Z&urlSigningMethod=HMACV1&urlSignature=jkgyMLkIyHtUCdhoIOt4QVJPkhaF2uKo8MZHFJMokEs%3D

tvdboom:
@ngupta23 We can install it through conda in the build. but we can't add it to the requirements file since those are installed through pip. I feel like the best options is:
 - remove prophet and the other mentioned packages from the requirements file.
 - add prophet and other package installation in the build to test specific code parts
 - add prophet dedicated section in the docs to inform the user they need to install it separately (and how) for specific functionalities

what do you think?

ngupta23:
> remove prophet and the other mentioned packages from the requirements file.
add prophet and other package installation in the build to test specific code parts

Yes, this should be ok as long as the unit tests exercise the Prophet model. I am ok leaving the actual install up to the user (handled by the user outside of pycaret).

> add prophet dedicated section in the docs to inform the user they need to install it separately (and how) for specific functionalities

I think when I add the improved messaging for optional packages, I can incorporate it there itself. If needed, we can update the docs as well.

tvdboom:
oke,requirements files are done then. could you have a look at how to add conda and prophet to the build @ngupta23? not sure if conda is already installed or not... if not, its gonna be a pain

ngupta23:
> if conda is already installed or not... if not, its gonna be a pain

@tvdboom I have not done it myself, but does this help - could you check?

https://autobencoder.com/2020-08-24-conda-actions/

ngupta23:
@tvdboom tried many things. None of them work. We will have to talk to @Yard1 unless you know of a better way.

ngupta23:
@tvdboom @Yard1 I think it is fixed now. It seems that the order of installation is important - not sure why but prophet has to be installed after the optional dependencies. The tests that are failing are only on the refactor branch and the failure is due to the multicollinearity threshold. I had already fixed this on the time series branch. 

I think we can merge `time_series` branch into this refactor branch to confirm or if we are comfortable, directly merge this branch into `time_series`.

ngupta23 (approved):

Yard1 (approved):
LGTM, thanks for your work guys!

Issue #2343, opened 2022-03-29
ACF and PACF plots to reuse subplot logic directly instead of replicating code.


ngupta23:
In Time series module, we can create ACF & PACF plots either 

1. Individually (on data or residuals)

```python
# On original data ----
plot_model(plot = "acf")

# On model residuals ----
plot_model(model, plot = "acf")
```

This uses the `plot_acf` and `plot_pacf` functions as shown below

https://github.com/pycaret/pycaret/blob/915a13d3b4d880b5a0dac0797465217b2ba934b9/pycaret/internal/plots/time_series.py#L446

https://github.com/pycaret/pycaret/blob/915a13d3b4d880b5a0dac0797465217b2ba934b9/pycaret/internal/plots/time_series.py#L539

OR 

2. As part of larger plots as subplots

```python
# On original data ----
plot_model(plot = "diagnostics")

# On model residuals ----
plot_model(model, plot = "diagnostics")
```

This uses the `corr_subplot` as shown below

https://github.com/pycaret/pycaret/blob/915a13d3b4d880b5a0dac0797465217b2ba934b9/pycaret/internal/plots/time_series.py#L783

Currently there is replication of code for these 2 use cases. The subplot approach is more modular (and has been cleaned up). Need to update `plot_acf` and `plot_pacf` to reuse the `corr_subplot` component instead of replicating the code to create these plots again.

ngupta23:
Note: When this change is made, `plot_acf` and `plot_pacf` will be essentially converted to a plotly subplot with 1 row and 1 column. This will allow us to use `corr_subplot` within `plot_acf` as follows (similarly for `plot_pacf`)

```python
fig, acf_data = corr_subplot(fig=fig, data=data, row=1, col=1, plot="acf") 
```

ryanxjhan:
Hi, you can assign this to me.

ngupta23:
@ryanxjhan , Awesome. I have assigned this to you. 

Please create a branch from `time_series` and submit the PR back to the `time_series` branch itself (not `master`)

PR #2346, opened 2022-03-30
Updated ACF and PACF plots to reuse subplot logic directly

ryanxjhan:
## Related Issues or Bug

Implemented #[2343](https://github.com/pycaret/pycaret/issues/2343)

Updated plot_acf and plot_pacf to reuse the corr_subplot component instead of replicating the code to create these plots again.

## Type of change

## How Has This Been Tested?

Tested locally with `pytest`

NA

## Checklist:

## Screenshots

 Original           | Updated

### DIFF
--- a/pycaret/internal/plots/time_series.py
+++ b/pycaret/internal/plots/time_series.py
@@ -459,7 +459,8 @@ def plot_acf(
     fig_defaults : Dict[str, Any]
         The defaults dictionary containing keys for "width" and "height" (mandatory)
     model_name : Optional[str]
-        The name of the model, by default None
+        The model_name must be passed to model residuals, while it should be left None
+        for the original data and the name of the data is passed
     data_kwargs : Dict[str, Any]
         A dictionary containing options keys for "nlags"
     fig_kwargs : Dict[str, Any]
@@ -474,11 +475,10 @@ def plot_acf(
     fig_kwargs = fig_kwargs or {}
 
     nlags = data_kwargs.get("nlags", None)
-    name = "ACF"
 
     subplots = make_subplots(rows=1, cols=1)
     fig, acf_data = corr_subplot(
-        fig=subplots, data=data, col=1, row=1, name=name, plot="acf", nlags=nlags
+        fig=subplots, data=data, col=1, row=1, plot="acf", nlags=nlags
     )
 
     time_series_name = data.name
@@ -490,16 +490,19 @@ def plot_acf(
 
     with fig.batch_update():
         fig.update_xaxes(title_text="Lags", row=1, col=1)
-        fig.update_yaxes(title_text=name, row=1, col=1)
+        fig.update_yaxes(title_text="ACF", row=1, col=1)
         template = _resolve_dict_keys(
             dict_=fig_kwargs, key="template", defaults=fig_defaults
         )
         fig.update_layout(title=title, showlegend=False, template=template)
+        fig.update_traces(marker={"size": 12})
         fig = _update_fig_dimensions(
             fig=fig, fig_kwargs=fig_kwargs, fig_defaults=fig_defaults
         )
 
-    return_data_dict = {"acf": acf(data, alpha=0.05, nlags=nlags)}
+    return_data_dict = {
+        "acf": (acf_data[0], np.column_stack((acf_data[2], acf_data[3])))
+    }
 
     return fig, return_data_dict
 
@@ -520,7 +523,8 @@ def plot_pacf(
     fig_defaults : Dict[str, Any]
         The defaults dictionary containing keys for "width" and "height" (mandatory)
     model_name : Optional[str]
-        The name of the model, by default None
+        The model_name must be passed to model residuals, while it should be left None
+        for the original data and the name of the data is passed
     data_kwargs : Dict[str, Any]
         A dictionary containing options keys for "nlags"
     fig_kwargs : Dict[str, Any]
@@ -535,11 +539,10 @@ def plot_pacf(
     fig_kwargs = fig_kwargs or {}
 
     nlags = data_kwargs.get("nlags", None)
-    name = "PACF"
 
     subplots = make_subplots(rows=1, cols=1)
-    fig, acf_data = corr_subplot(
-        fig=subplots, data=data, col=1, row=1, name=name, plot="pacf", nlags=nlags
+    fig, pacf_data = corr_subplot(
+        fig=subplots, data=data, col=1, row=1, plot="pacf", nlags=nlags
     )
 
     time_series_name = data.name
@@ -551,16 +554,19 @@ def plot_pacf(
 
     with fig.batch_update():
         fig.update_xaxes(title_text="Lags", row=1, col=1)
-        fig.update_yaxes(title_text=name, row=1, col=1)
+        fig.update_yaxes(title_text="PACF", row=1, col=1)
         template = _resolve_dict_keys(
             dict_=fig_kwargs, key="template", defaults=fig_defaults
         )
         fig.update_layout(title=title, showlegend=False, template=template)
+        fig.update_traces(marker={"size": 12})
         fig = _update_fig_dimensions(
             fig=fig, fig_kwargs=fig_kwargs, fig_defaults=fig_defaults
         )
 
-    return_data_dict = {"pacf": pacf(data, alpha=0.05, nlags=nlags)}
+    return_data_dict = {
+        "pacf": (pacf_data[0], np.column_stack((pacf_data[2], pacf_data[3])))
+    }
 
     return fig, return_data_dict
 
--- a/pycaret/internal/plots/time_series.py
+++ b/pycaret/internal/plots/time_series.py
@@ -450,14 +450,36 @@ def plot_acf(
     data_kwargs: Optional[Dict] = None,
     fig_kwargs: Optional[Dict] = None,
 ) -> PlotReturnType:
-    """Plots the ACF on the data provided"""
-    fig, return_data_dict = None, None
+    """Plots the ACF on the data provided
 
+    Parameters
+    ----------
+    data : pd.Series
+        Data whose correlation plot needs to be plotted
+    fig_defaults : Dict[str, Any]
+        The defaults dictionary containing keys for "width" and "height" (mandatory)
+    model_name : Optional[str]
+        The name of the model, by default None
+    data_kwargs : Dict[str, Any]
+        A dictionary containing options keys for "nlags"
+    fig_kwargs : Dict[str, Any]
+        A dictionary containing options keys for "width" and/or "height"
+
+    Returns
+    -------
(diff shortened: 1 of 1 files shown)


pycaret/internal/plots/time_series.py
```diff
@@ -529,9 +499,7 @@ def plot_acf(
             fig=fig, fig_kwargs=fig_kwargs, fig_defaults=fig_defaults
         )
 
-    return_data_dict = {
-        "acf": corr_array,
-    }
+    return_data_dict = {"acf": acf(data, alpha=0.05, nlags=nlags)}
```
ngupta23, line 502: You should reuse the `acf_data` returned from the subplot above here instead of calling acf function again.

pycaret/internal/plots/time_series.py
```diff
@@ -561,74 +549,18 @@ def plot_pacf(
...
         fig.update_layout(title=title, showlegend=False, template=template)
-
         fig = _update_fig_dimensions(
             fig=fig, fig_kwargs=fig_kwargs, fig_defaults=fig_defaults
         )
 
-    return_data_dict = {
-        "pacf": corr_array,
-    }
+    return_data_dict = {"pacf": pacf(data, alpha=0.05, nlags=nlags)}
```
ngupta23, line 563: You should reuse the `pacf_data` returned from the subplot above here instead of calling pacf function again.

pycaret/internal/plots/time_series.py
```diff
@@ -543,16 +511,36 @@ def plot_pacf(
...
 
-    nobs = len(data)
-    nlags_default = min(int(10 * np.log10(nobs)), nobs // 2 - 1)
-    nlags = data_kwargs.get("nlags", nlags_default)
-    corr_array = pacf(data, nlags=nlags, alpha=0.05)
+    nlags = data_kwargs.get("nlags", None)
+    name = "PACF"
+
+    subplots = make_subplots(rows=1, cols=1)
+    fig, acf_data = corr_subplot(
```
ngupta23, line 541: renames this to `pacf_data` instead of `acf_data`.

pycaret/internal/plots/time_series.py
```diff
@@ -450,14 +450,36 @@ def plot_acf(
...
-    fig, return_data_dict = None, None
-
+    """Plots the ACF on the data provided
+
+    Parameters
+    ----------
+    data : pd.Series
+        Data whose correlation plot needs to be plotted
+    fig_defaults : Dict[str, Any]
+        The defaults dictionary containing keys for "width" and "height" (mandatory)
```
ngupta23, line 460: Thanks for adding the docstring in the right numpy format. I have been meaning to do that for a while now.

I don't think these are mandatory. Plus it may not necessarily be limited to "width" and "height". You could also have template, etc. I would just say, something like "The defaults settings for the plotly figure."

pycaret/internal/plots/time_series.py
```diff
@@ -450,14 +450,36 @@ def plot_acf(
...
+    model_name : Optional[str]
+        The name of the model, by default None
```
ngupta23, line 462: I would clarify here, If ACF is for model residuals, model_name must be passed. If model_name is `None`, then name is derived from `data`.

pycaret/internal/plots/time_series.py
```diff
@@ -450,14 +450,36 @@ def plot_acf(
...
     data_kwargs = data_kwargs or {}
     fig_kwargs = fig_kwargs or {}
 
     nlags = data_kwargs.get("nlags", None)
-    corr_array = acf(data, alpha=0.05, nlags=nlags)
+    name = "ACF"
+
+    subplots = make_subplots(rows=1, cols=1)
+    fig, acf_data = corr_subplot(
+        fig=subplots, data=data, col=1, row=1, name=name, plot="acf", nlags=nlags
```
ngupta23, line 481: You do not have to pass name = name here since it defaults to "ACF". If you are changing it, then you can pass it.

pycaret/internal/plots/time_series.py
```diff
@@ -466,61 +488,9 @@ def plot_acf(
...
-        line_color="#3f3f3f",
-        name="",
-    )
-
     with fig.batch_update():
-        fig.update_xaxes(range=[-1, len(corr_array[0]) + 1])
-        fig.update_yaxes(zerolinecolor="#000000")
-
+        fig.update_xaxes(title_text="Lags", row=1, col=1)
+        fig.update_yaxes(title_text=name, row=1, col=1)
```
ngupta23, line 493: Directly use "ACF" here instead of `name` to make it clear, especially, if you remove name=name from the `corr_subplot` above.

pycaret/internal/plots/time_series.py
```diff
@@ -543,16 +511,36 @@ def plot_pacf(
     data_kwargs: Optional[Dict] = None,
     fig_kwargs: Optional[Dict] = None,
 ) -> PlotReturnType:
-    """Plots the PACF on the data provided"""
-    fig, return_data_dict = None, None
-
+    """Plots the PACF on the data provided
```
ngupta23, line 514: Similar changes as the `plot_acf` should be made in `plot_pacf` as well. Thanks!

ngupta23:
Hi @ryanxjhan 

I have one more request. When we plot ACF as part of subplots, we reduce the size of the stem endpoints to 6 (from 10) for a more compact display. But in this case, it makes the individual ACF plot endpoint look very small as well.

Previously

image

With these changes

image

I am wondering if this could be customized in the code so, in the call from `plot_acf` and `plot_pacf`, we could use 10 and leave it as 6 by default for other plots. Let me know what you think?

image

ngupta23:
@ryanxjhan  Don't worry about the failing test. It is unrelated to time series plotting.

ryanxjhan:
@ngupta23 Thank you for the feedback. I implemented all the changes above successfully. Tested it with Jupyter and pytest.

pycaret/internal/plots/time_series.py
ngupta23: Any reason why you want to pass name = name here. The default is already set to ACF for this plot internally.

pycaret/internal/plots/time_series.py
```diff
@@ -450,14 +450,36 @@ def plot_acf(
...
+    """Plots the ACF on the data provided
+
+    Parameters
+    ----------
+    data : pd.Series
+        Data whose correlation plot needs to be plotted
+    fig_defaults : Dict[str, Any]
+        The defaults dictionary containing keys for "width" and "height" (mandatory)
+    model_name : Optional[str]
+        The model_name must be passed to model residuals, while it should be left None
```
ngupta23, line 462: The wording here is confusing. Could you please change it to the following.

> If the correlation plot is for model residuals, then, model_name must be passed for proper display of results. If the correlation plot is for the original data, model_name should be left None (name is derived from the data passed in this case).

pycaret/internal/plots/time_series.py
```diff
@@ -466,71 +488,20 @@ def plot_acf(
...
-        fig.update_xaxes(range=[-1, len(corr_array[0]) + 1])
-        fig.update_yaxes(zerolinecolor="#000000")
-
+        fig.update_xaxes(title_text="Lags", row=1, col=1)
+        fig.update_yaxes(title_text="ACF", row=1, col=1)
         template = _resolve_dict_keys(
             dict_=fig_kwargs, key="template", defaults=fig_defaults
         )
         fig.update_layout(title=title, showlegend=False, template=template)
+        fig.update_traces(marker={"size": 12})
```
ngupta23, line 498: Can you change the size to 10 (not 12) in order to be consistent with the previous version? Thanks!

pycaret/internal/plots/time_series.py
```diff
@@ -543,16 +514,36 @@ def plot_pacf(
...
+    """Plots the PACF on the data provided
+
+    Parameters
+    ----------
+    data : pd.Series
+        Data whose correlation plot needs to be plotted
+    fig_defaults : Dict[str, Any]
+        The defaults dictionary containing keys for "width" and "height" (mandatory)
+    model_name : Optional[str]
+        The model_name must be passed to model residuals, while it should be left None
```
ngupta23, line 526: Similar to ACF plot, the wording here is confusing. Could you please change it to the following.

> If the correlation plot is for model residuals, then, model_name must be passed for proper display of results. If the correlation plot is for the original data, model_name should be left None (name is derived from the data passed in this case).

pycaret/internal/plots/time_series.py
```diff
@@ -466,71 +488,20 @@ def plot_acf(
...
         )
         fig.update_layout(title=title, showlegend=False, template=template)
+        fig.update_traces(marker={"size": 12})
         fig = _update_fig_dimensions(
             fig=fig, fig_kwargs=fig_kwargs, fig_defaults=fig_defaults
         )
 
     return_data_dict = {
-        "acf": corr_array,
+        "acf": (acf_data[0], np.column_stack((acf_data[2], acf_data[3])))
```
ngupta23, line 504: For now, can you just return `acf_data` as is (in order to be consistent with what we do in other plots like diagnostics)? If the formatting needs to change, we can change it later in the underlying `corr_subplot` function (that can be a separate PR after discussing what the format needs to be).

```python
return_data_dict = {"acf": acf_data}
```

pycaret/internal/plots/time_series.py
```diff
@@ -561,73 +552,20 @@ def plot_pacf(
...
-        fig.update_yaxes(zerolinecolor="#000000")
-
+        fig.update_xaxes(title_text="Lags", row=1, col=1)
+        fig.update_yaxes(title_text="PACF", row=1, col=1)
         template = _resolve_dict_keys(
             dict_=fig_kwargs, key="template", defaults=fig_defaults
         )
         fig.update_layout(title=title, showlegend=False, template=template)
-
+        fig.update_traces(marker={"size": 12})
```
ngupta23, line 562: Similar to ACF, please make this 10

pycaret/internal/plots/time_series.py
```diff
@@ -561,73 +552,20 @@ def plot_pacf(
...
         fig = _update_fig_dimensions(
             fig=fig, fig_kwargs=fig_kwargs, fig_defaults=fig_defaults
         )
 
     return_data_dict = {
-        "pacf": corr_array,
+        "pacf": (pacf_data[0], np.column_stack((pacf_data[2], pacf_data[3])))
```
ngupta23, line 568: Similar to ACF, return data as is

```python
return_data_dict = {"pacf": pacf_data}
```

ngupta23 (changes requested):
Hi @ryanxjhan , Thanks for making the changes. I have some minor feedback. Once this is done, this should be good to go. Thanks!

ryanxjhan:
Hi @ngupta23, thank you! The changes have been made.

PR #2410, opened 2022-04-14
AJarman/issue1766

### LINKED ISSUE #1766: Add lower limit and upper limit for time series forecasting

AJarman:
## Related Issue or bug

> Hi @AJarman
> 
> I discussed this with @moezali1 and @Yard1 in the meeting on 20220409 and this is the recommendation. We should add an argument to setup called `forecast_limit: Optional[List[Union[int, float]] = None`. By default, no limits are applied (same as today's functionality). If user wants to set limits, they can do it as follows:
> 
> ```python
> forecast_limit = None # default - no limits
> forecast_limit = [0, 10000000] # lower and upper limit
> forecast_limit = [0, None]  # lower limit only
> forecast_limit = [None, 10000000] # upper limit only
> ```
> 
> ## Coding Actions:
> * Inside `setup`, there will have to be a check to make sure that if a list is passed, it has exactly 2 entries and the types are correct.
> * If this is enabled, we should add a step to the transformation pipeline. The order of transformations will be as follows
>   
>   * Impute
>   * Limit
>   * Transform
>   * Scale
> 
> ## Future enhancements and user customization
> In the future, we will also extend the following arguments to accept Transformer objects directly so any user wanting to set custom arguments, can pass the object directly (this is similar to what the other modules such as regression and classification do in pycaret). This will do away with the need for the kwargs in setup.
> 
> ```python
> transform_target = "box_cox"  # Today
> transform_target = BoxCoxTransformer(limits = Sth, method="pearsonr")  # In the future (in lieu of transform_kwargs)
> 
> forecast_limit = [0, None] # current proposal
> forecast_limit = ScaledLogitTransformer(0, None) # In the future (in lieu of transform_kwargs)
> ```
> 
> Let me know if this makes sense. Would be happy to discuss more if needed.
> 
> Thanks!

Closes #1766

I have implemented the sktime `scaledLogitTransformer` in the time series module. This is a pipeline step added after imputatation, before transformation and scaling. This currently only affects the target variable and is specified using a list of two values.
Implementing the same for exogenous variables would also be straightforward, however at the moment I'm not sure how this would function in practice, given that they would presumably need different limits.
I have left a 'stubbed' out option (elif) which would be used to implement passing an sktime estimator directly as specified in the future requirements, at present this raises a specific `NotImplementedError`, rather than the generic `TypeError` it would do for any other type.

## Type of change

## How Has This Been Tested?

Tests TBC - in progress

## Checklist:

## Screenshots

 Original           | Updated


AJarman:
Note still in draft, demo notebook and unit tests to follow

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
```diff
@@ -112,6 +114,101 @@ def _add_imputation_steps(
         else:
             self.pipe_steps_exogenous.extend([("numerical_imputer", num_estimator)])
 
+    def _limitation(self,
```
ngupta23, line 117: Could you rename this to _limiter? _limitation sounds like some limitation of the module itself. Sorry just being picky :)

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
```diff
@@ -112,6 +114,101 @@ def _add_imputation_steps(
...
+                    limit_target: Optional[List[Union[int,float,None]]],
+                    limit_exogenous: Optional[List[Union[int,float,None]]],
+                    exogenous_present: bool
+                    ):
+
+     #### Limit target ----
+        if limit_target is not None:
+            self._add_limitation_steps(
```
ngupta23, line 125: Same as above - maybe call this `_add_limiter_step`

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
```diff
@@ -112,6 +114,101 @@ def _add_imputation_steps(
...
+        # TODO: Not yet implemented
+        # Will raise an error
+        # Only add exogenous pipeline steps if exogenous variables are present.
+        if (
+            exogenous_present == TSExogenousPresent.YES
+            and limit_exogenous is not None
+        ):
+            # self._add_limitation_steps(
+            #     limits=limit_exogenous, target=False
+            self.logger.warning("Applying limits to exogenous variables is not yet implemented.")
```
ngupta23, line 138: When will this condition be triggered by user inputs?

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
```diff
@@ -112,6 +114,101 @@ def _add_imputation_steps(
...
+                "Using transformers directly is not yet implemented,\
+                    please use numeric limits only for now.")
+        else:
+            raise TypeError(
+                f"{type_} forecast limit Type '{type(limits)}' is not of allowed type."
+            )
+
+        if target:
+            print("creating pipeline limit step")
+            self.pipe_steps_target.extend([("target_limiter", limiter)])
```
ngupta23, line 206: I would just call it limiter. No need to explicitly call out `target` since this will go inside `TransformedTargetForecaster` and will be different from the steps that go into the `ForecastingPipeline` for exogenous variables.

pycaret/tests/test_time_series_preprocess.py
```diff
@@ -600,3 +600,17 @@ def test_pipeline_after_finalizing(load_pos_and_neg_data_missing):
...
+
+    exp = TSForecastingExperiment()
+    FH = 12
+    lower_bound = 0.0
+    exp.setup(data=data, fh=FH, limit_target=[lower_bound, None], preprocess=True)
+
+    # Currently fails due to missing values
+    assert np.all(exp.y_transformed.values >= lower_bound)
+    assert np.all(exp.y_train_transformed.values >= lower_bound)
+    assert np.all(exp.y_test_transformed.values >= lower_bound)
```
ngupta23, line 616: The tests should check that the forecast values are within the limits. So, you could create an artificial dataset with a declining trend that will forecast below the lower limit for example. Then run once with a limiter and once without the limiter. In the former case, it would have predictions below the lower limit and in the latter case, it would not have any predictions below the lower limit.

Then repeat the same for the upper limit.

pycaret/time_series/forecasting/oop.py
```diff
@@ -1242,6 +1248,15 @@ def setup(
                 "bfill", "pad", "ffill", "random"
             If int or float, imputation method is set to "constant" with the given value.
 
+        limit_target: Optional[List[Optional[float]]], default = None
```
ngupta23, line 1251: You will need to add the same argument to the functional API py file as well (in the same folder as oop.py)

ngupta23:
Hi @AJarman I have left my comments on the PR. They are all minor. I think the feedback here will help resolve these issues:

https://github.com/pycaret/pycaret/issues/1766#issuecomment-1101601002

Issue #2364, opened 2022-04-03
[BUG] Time Series Plotting when data is transformed is not good.


ngupta23:
**Describe the bug**

If data has missing values, and the user decides to impute the data, the plotting still works on the original dataset (which has missing values). This causes a lot of plots to not display any results.

Moreover, some plots like decomposition fail since we have missing values.

**To Reproduce**

```python
import numpy as np
import pandas as pd
from pycaret.datasets import get_data
from pycaret.time_series import TSForecastingExperiment

data = get_data("airquality")
data["index"] = pd.to_datetime(data["Date"] + " " + data["Time"])
data.drop(columns=["Date", "Time"], inplace=True)
target = "CO(GT)"

data.replace(-200, np.nan, inplace=True)
# data.fillna(method="ffill", inplace=True)

#### Create an EDA experiment ----
eda = TSForecastingExperiment()

eda.setup(
    data=data,
    target=target,
    index="index",
    fh=48,
    numeric_imputation_target="ffill",
    numeric_imputation_exogenous="ffill",
    # Set defaults for the plots ----
    # fig_kwargs={"renderer": "notebook", "width": 1000, "height": 600},
    session_id=42,
)

# Does not plot well with missing values
eda.plot_model(plot="diagnostics", fig_kwargs={"renderer": "browser"})

# This fails.
eda.plot_model(plot="decomp", fig_kwargs={"renderer": "browser"})
```

![image](https://user-images.githubusercontent.com/33585645/161437473-6e02f76f-d3b5-4743-9f85-1e9a5362d1f1.png)

**Expected behavior**
Maybe plots should work on the transformed data (by default) with the option for the user to use the original data for plotting if needed.

**Versions**
Dev version as of 3rd April 2022

## Plots to fix

- [x] ts https://github.com/pycaret/pycaret/commit/76bcc2270e4bda92ec9019c20e5e08598e3da40b
- [x] train_test_split https://github.com/pycaret/pycaret/commit/76bcc2270e4bda92ec9019c20e5e08598e3da40b
- [ ] cv (will do later after 3.0 release)
- [x] acf https://github.com/pycaret/pycaret/pull/2416/commits/e08aafdeda07268ca867a164c8eb41c67dcce87a
- [x] pacf https://github.com/pycaret/pycaret/pull/2416/commits/e08aafdeda07268ca867a164c8eb41c67dcce87a
[...]

PR #2413, opened 2022-04-15
Missing Data plot fixes

ngupta23:
## Related Issuse or bug

- Partially Implements #2364
- Also fixed a bug in the retrieval of test values

- Added ability for `ts` and `train_test_split` plots to accept different plot data types - "original", "imputed", "transformed"
- Also fixed a bug in the retrieval of test values - these should be computed by passing the entire dataset through the pipeline and then retrieving the transformed test set values instead of just the test set through the pipeline (since we can use all past data but no future data)dataset values

## Type of change

## How Has This Been Tested?

Unit tests have been added for plotting as well as pipeline

## Checklist:

### DIFF
--- a/pycaret/internal/plots/utils/time_series.py
+++ b/pycaret/internal/plots/utils/time_series.py
@@ -1,5 +1,6 @@
 from typing import Optional, List, Tuple, Union, Dict, Any
 from math import ceil
+import logging
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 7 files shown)


pycaret/internal/plots/utils/time_series.py
```diff
@@ -24,6 +25,46 @@
 PlotReturnType = Tuple[Optional[go.Figure], Optional[Dict[str, Any]]]
 
 
+#### Data Types allowed for each plot type ----
+# First one in the list is the default (if requested is None)
+ALLOWED_PLOT_DATA_TYPES = {
```
Yard1, line 30: Maybe we can use enums here?

Yard1 (commented):
Logic looks good, but I think we whouls be using enums for data types and split types instead of regular strings (ofc we should allow the user to pass in a string, but we should use enums in the code). Let me know what you think

pycaret/internal/plots/utils/time_series.py
ngupta23: ok, makes sense. Maybe I will add that as a final refactor to make it easier since I am in the middle of more changes. Hope that is fine.

ngupta23: I opened another issue for this for tracking. Does this only apply to ALLOWED_PLOT_DATA_TYPES? 
Also, are you ok merging this if we do this refactor later?

Issue #2618, opened 2022-06-02
[BUG]: finalize_model might be causing issues


moezali1:
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).

- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.

- [X] I have confirmed this bug exists on the develop branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@develop).

### Issue Description

I am running REG 101 tutorial off the `develop` branch and I noticed something really strange in terms of performance. This wasn't happening before. 

There is a sudden performance drop when we `finalize_model`. This wasn't happening before. Infact I expect performance to go up especially for tree based models when `finalize_model` is used - for obvious reasons.

but in this case 97% R2 goes to 60's after `finalize_model`. Do you mind running Official REG101 tutorial on `develop` branch and ensure there is nothing wrong here.

![image](https://user-images.githubusercontent.com/54699234/171628748-08b835a4-1f5a-4afc-b30f-e2ef18fe0b0f.png)

### Reproducible Example

```python
.
```

### Expected Behavior

no performance drop

### Actual Results

```python-traceback
.
```

### Installed Versions

3.0.0-rc

PR #2647, opened 2022-06-08
fix finalize_model

tvdboom:
## Related Issue or bug

Large refactor of `finalize_model` and some minor bug fixes.

About `finalize_model`:

I had a good look at the `finalize_model` method after @moezali1 opened #2618. The problem with the old implementation is that it only fitted the model and not the pipeline. This doesn't make sense since the use case for this method is to do a final training on all available data to use the resulting model in production. But you can't use the model without the transformation pipeline, and you need to fit the pipeline on the same data as the model. Therefore, the new implementation always fits the whole pipeline and returns the pipeline or the model depending on the parameter `model_only`, which now defaults to False.

Once you have the whole pipeline fitted, calling `predict_model` directly like this `predict_model(pipeline)` will yield poor results, and rightfully so since `predict_model` uses the transformed test set (transformed on the training set only, a different data set than the pipeline). To get the expected results (near perfect scores), one needs to call `predict_model(pipeline, data=original_dataset)` because in this case the method will correctly transform the provided data. See the image below running the regression tutorial.

Let me know if you disagree with this or have any questions.

@ngupta23 I wasn't able to reconcile this new implementation with time series because of the results are retrieved for logging. Could you have a look if we decide that this is the right way? thanks

Enhancements:

* All tests ran now in a temporary directory, avoiding all the files in the home dir. Solves #2648 

Bug fixes:

* The `X_test_transformed` and `y_test_transformed` attributes need to be transformed only on the data (X or y) they represent to avoid issues with category encoders (see [here](https://contrib.scikit-learn.org/category_encoders/leaveoneout.html#category_encoders.leave_one_out.LeaveOneOutEncoder.transform) for r example)
* `predict_model` didn't copy the pipeline if passed and removed the final estimator, therefore also removing the estimator from the original object

## Type of change

## Screenshots

temp


Yard1 (commented):
Thanks, this looks great. Let's discuss returning the pipeline by default - I think that's fine, but interested to hear what @moezali1 and @ngupta23 have to say.

CI seems to be failing

moezali1:
@tvdboom I have one question. If a pipeline is fitted on entire dataset and I use `predict_model(final_pipeline)` why it should have bad score? `X_test_transformed` is used by default in `predict_model` and when the pipeline is fitted on entire dataset `X_test_transformed` is part of it. So shouldn't we get near perfect scores here as well?

tvdboom:
@moezali1 'X_test_transformed' has been transformed using a different pipeline then final_pipeline. Therefore the data it contains is not the same as 'final_pipeline.transform(X_test)', which is what is needed for a perfect score

ngupta23:
@tvdboom I looked at this briefly and I think the current implementation on the `develop` branch for `finalize_model` with `model_only=False` is not correct for time series. You may have to do something like this (like done in `predict_model` in time series)

image

Basically, in time series, there are 2 pipelines (one that has the memory of the training dataset only and one that has the memory of the full dataset (we can do this in time series since the model stores the data used for training as it is auto-regressive). Then we pick the right pipeline to append depending on the data in the final model.

pycaret/internal/pycaret_experiment/supervised_experiment.py
```diff
@@ -4654,58 +4661,56 @@ def finalize_model(
...
-            groups=groups,
+            fit_kwargs=fit_kwargs or {},
+            predict=False,
+            groups=self._get_groups(groups, data=self.X),
             add_to_model_list=False,
-            return_train_score=return_train_score,
-            # TODO fix this, finalize should work without CV
-            cross_validation=True,
+            model_only=False,
+            return_train_score=False,
```
Yard1, line 4686: should we remove `return_train_score` from args of the method?

pycaret/internal/pycaret_experiment/supervised_experiment.py
```diff
@@ -4716,18 +4721,20 @@ def finalize_model(
...
-            pipeline_final = deepcopy(self.pipeline)
-            pipeline_final.steps.append(["trained_model", model_final])
-            return pipeline_final
 
-        return model_final
+        if model_only:
+            if self._ml_usecase != MLUsecase.TIME_SERIES:
+                return pipeline_final._final_estimator
+            else:
+                return self._get_final_model_from_pipeline(pipeline_final)
```
Yard1, line 4735: can we make this a method in supervised experiment and call it here as well, so we don't have a special case, just method overriding?

Yard1 (approved):
LGTM, thanks a lot for working on this! Some small nits

Ali-Flt:
I wanted to check if `finalize_model()` is actually fixed now and if yes a bit of explanation on how its results should be evaluated and interpreted.
As mentioned before, I have the same issue where the model before `finalize_model()` has a `R2 = 0.75` on test data when running `predict_model()` but after `finalize_model()` the R2 goes to minus values. But interestingly when I do `predict_model(data=all_data)` where `all_data` is all of my dataset given to `setup()` function (train + test together) I get a really good R2 and the prediction is really good even for the same rows which comprise the test_data.
Therefore, I'm not quite sure whether there's something wrong with `finalize_model()` or I'm just doing something wrong.
```
Pycaret version: 3.0.0rc3
```

tvdboom:
In the PR description is explainer why you are getting the results you mention.

> Once you have the whole pipeline fitted, calling predict_model directly like this predict_model(pipeline) will yield poor results, and rightfully so since predict_model uses the transformed test set (transformed on the training set only, a different data set than the pipeline). To get the expected results (near perfect scores), one needs to call predict_model(pipeline, data=original_dataset) because in this case the method will correctly transform the provided data. See the image below running the regression tutorial.

`finalize_model` is intended to use **all** the available data to train the model. This means there should be no data left for evaluating the final model. If you need to evaluate your model, don't use `finalize_model`, but theoretically, the evaluation has been done already before.

Issue #2784, opened 2022-07-20
[BUG]: Datetime extraction features drop time


celestinoxp:
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).

- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.

- [X] I have confirmed this bug exists on the master branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@master).

### Issue Description

I have a dataframe with a feature data_time. after setup, pycaret have created features for year, month and day, however have droped the time (hour, minutes, seconds)

2022-01-01 20:00:00

### Reproducible Example

```python
.
```

### Expected Behavior

.

### Actual Results

```python-traceback
.
```

### Installed Versions

<details>
rc3
</details>

tvdboom:
The PR above adds a new parameter `create_date_columns` where the user can choose which features to extract from `date_features`. It will be available in `master` after the merge.

celestinoxp:
> 

Can convert example 2022-01-01 20:00:00
year: 2022
month: 01
day: 01
hour: 20
minutes: 00
seconds: 00

?

tvdboom:
You have to select them yourself using create_date_columns=["year", "month", "day", "hours", "minutes", "seconds"]

PR #2798, opened 2022-07-25
Add `create_date_columns`, replace column names in `predict_model`

### LINKED ISSUE #2790: [BUG]: Prediction column named the same as target column if target == "Label"
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).

- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.

- [ ] I have confirmed this bug exists on the master branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@master).

### Issue Description

In my dataset, the target-column is named "Label". Thus, when predicting on a model it will create a new column named "Label". 

image

This makes it hard for me to use that dataframe, since I do not know which column is my target-column or which is the predicted column. 

This has lead to tedious workaround in my code, changing column names to "_Label". I think these workaround are unnecessary and could be handled by PyCaret instead.

Thanks! 

### Reproducible Example

```python
df = get_data("juice", verbose=False)
df = df.rename(columns={"STORE":"Label"})
exp = ClassificationExperiment()
exp.setup(data=df)

clf = exp.create_model("rf")

predictions = exp.predict_model(clf)
predictions
```

### Expected Behavior

Maybe change the name to "predicted_label" or similar.

Or check if target == "Label", change predicted column to "Label_1"

### Actual Results

```python-traceback
See picture above.
```

### Installed Versions

<details>
PyCaret 3.0.0rc3
</details>

tvdboom:
Closes #2790 and #2784

* It's now possible to choose which datetime columns to extract from `date_features` using parameter `create_date_columns` in setup.
 * In `predict_model`, the Label and Score column names have been replaced with `prediction_label` and `prediction_score`.

## Type of change

### DIFF
--- a/pycaret/internal/pycaret_experiment/supervised_experiment.py
+++ b/pycaret/internal/pycaret_experiment/supervised_experiment.py
@@ -5222,7 +5222,7 @@ class _SupervisedExperiment(_TabularExperiment):
         if self._ml_usecase == MLUsecase.CLASSIFICATION:
             metric_dict["Selection Rate"] = selection_rate
 
-        y_pred = self.predict_model(estimator)["Label"]
+        y_pred = self.predict_model(estimator)["prediction_label"]
         y_true = self.y_test
 
         try:
--- a/pycaret/internal/pycaret_experiment/supervised_experiment.py
+++ b/pycaret/internal/pycaret_experiment/supervised_experiment.py
@@ -60,6 +60,7 @@ from pycaret.internal.utils import (
 )
 from pycaret.internal.validation import is_fitted, is_sklearn_cv_generator
 from pycaret.utils._dependencies import _check_soft_dependencies
+from pycaret.utils.constants import LABEL_COLUMN, SCORE_COLUMN
 
 LOGGER = get_logger()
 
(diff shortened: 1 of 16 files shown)


pycaret/internal/pycaret_experiment/supervised_experiment.py
```diff
@@ -5029,15 +5029,17 @@ def replace_labels_in_column(pipeline, labels: pd.Series) -> pd.Series:
             df_score = df_score.round(round)
             display.display(df_score.style.format(precision=round))
 
-        label = pd.DataFrame(pred, columns=["Label"], index=X_test_.index)
+        label = pd.DataFrame(pred, columns=["prediction_label"], index=X_test_.index)
```
Yard1, line 5032: Actually, could we make a switch to a constant defined somewhere instead of hardcoding strings here and in other places? That would make it more maintainable

Yard1 (changes requested):

pycaret/internal/pycaret_experiment/supervised_experiment.py
tvdboom: would utils be the place for it or a new file called constants? not sure what best practices dictate here...

Yard1: Let's make `pycaret/constants.py`

ngupta23: Can you add constants.py under utils submodule to keep top level relatively clean?

Yard1: That would work, good idea!

Yard1 (approved):

PR #2826, opened 2022-08-05
added index handling

### LINKED ISSUE #2815: [ENH]: Parameter for index column in setup

Often, we use datasets which have different index columns than usual counter (0,1,2,...). For example a long identification string. In PyCaret, there is no option for setting this index column. 

Before passing in the dataframe, I can however manually set my index_column with `df.set_index("index_col")`. But this is not saved internally in PyCaret. 

Below you will find a bit messy example, creating a new index column for a standard dataset.

> import pandas as pd
import numpy as np
df = get_data("juice", verbose=False)
size = len(df)
ind_col = ["Instance " + str(i) for i in range(size)]
df_ind = pd.DataFrame(ind_col, columns=["index_col"])
df = pd.concat([df_ind, df],axis=1)
df = df.set_index("index_col")
exp = ClassificationExperiment()
exp.setup(df, target="STORE", verbose=False)
clf = exp.create_model("rf", verbose=False)

If I then write `X_train = exp.get_config("X_train")`, I will not see my index column in the `X_train`.

This also affects `predict_model()`, since the "index_col" in the unseen dataset is just dropped.

### Describe your proposed solution

Add a parameter in `setup()` for setting an index column. This can be stored internally and when doing data preprocessing (for example with an unseen dataset), the index column can be set automatically. 

`df = df.set_index(<stored_index_column>)`

_No response_

### Additional context

_No response_

tvdboom:
## Related Issue or bug

Closes #2815

Support custom indices for the dataset. The user can now specify the `index` parameter in `setup` to specify the indices for his dataset. If specified, the indices are not reset.

Additionally, the `fix_imbalance_method` parameter now also accepts any of the imblearn estimators as string.

## Type of change

### DIFF
--- a/pycaret/anomaly/functional.py
+++ b/pycaret/anomaly/functional.py
@@ -6,8 +6,8 @@ from joblib.memory import Memory
 
 from pycaret.anomaly.oop import AnomalyExperiment
 from pycaret.internal.utils import check_if_global_is_not_none
-from pycaret.utils.constants import SEQUENCE_LIKE, DATAFRAME_LIKE
 from pycaret.loggers.base_logger import BaseLogger
+from pycaret.utils.constants import DATAFRAME_LIKE, SEQUENCE_LIKE
 
 _EXPERIMENT_CLASS = AnomalyExperiment
 _CURRENT_EXPERIMENT: Optional[AnomalyExperiment] = None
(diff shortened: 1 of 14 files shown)


pycaret/anomaly/functional.py
```diff
@@ -95,6 +97,14 @@ def setup(
         names.
 
 
+    index: bool, int, str or sequence, default=False
```
Yard1, line 100: can we add a one sentence description of the argument and what it does? It's not clear to me from the bullet points. Something like "How to handle index in the `data` dataframe"

Yard1 (approved):
Thanks, looks great. one nit

PR #3000, opened 2022-09-28
removing extra auc plot checking

qubiit:
# Related Issue or bug

Check estimator’s “predict_proba” attribute for “auc” was implemented two times.

Check estimator’s “predict_proba” attribute for “auc” was implemented two times. One was deleted. 

# Type of change

# How Has This Been Tested?

# Describe if there is any unusual behavior of your code(Write `NA` if there isn't)

NA


Yard1 (approved):
Thanks!

PR #3013, opened 2022-10-04
Update requirements.txt

celestinoxp:
The package joblib from 0 and before 1.2.0 is vulnerable to Arbitrary Code Execution (https://github.com/advisories/GHSA-6hrg-qmvc-2xh8)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ numpy>=1.21, <1.23  # Can't >=1.23 because of sktime/numba
 pandas>=1.3.0, <1.6.0  # Can't >=1.6 because of sktime
 jinja2>=1.2  # Required by pycaret.internal.utils --> pandas.io.formats.style
 scipy<1.9.0  # Can't >=1.9.0 due to sktime
-joblib>=1.1.0
+joblib>=1.2.0 #The package joblib from 0 and before 1.2.0 is vulnerable to Arbitrary Code Execution (https://github.com/advisories/GHSA-6hrg-qmvc-2xh8)
 scikit-learn>=1.0
 pyod>=0.9.8
 imbalanced-learn>=0.8.1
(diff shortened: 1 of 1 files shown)


Yard1 (approved):
Thanks, looks good - let's wait for CI to pass and I'll merge

ngupta23:
In this case we are limited by pandas-profiling. Once they relax their joblib dependency requirement, we can update the one on our end.

image

celestinoxp:
> In this case we are limited by pandas-profiling. Once they relax their joblib dependency requirement, we can update the one on our end.
> 
> image

I already related the problem and it has already been solved. It only remains to wait for the new version to be released
See: https://github.com/ydataai/pandas-profiling/issues/1056

ngupta23:
> I already related the problem and it has already been solved. It only remains to wait for the new version to be released

Excellent. Will wait for their release to merge this request.

celestinoxp:
> > I already related the problem and it has already been solved. It only remains to wait for the new version to be released
> 
> Excellent. Will wait for their release to merge this request.

I spoke with a member of pandas-profiling to try to resolve this issue quickly.
I proposed to release an immediate interim release, but my request was not approved. It looks like we're going to have to wait a week or two... even though there's a security issue regarding joblib with version lower than 1.2.

requirements.txt
```diff
@@ -6,7 +6,7 @@ numpy>=1.21, <1.23  # Can't >=1.23 because of sktime/numba
 pandas>=1.3.0, <1.6.0  # Can't >=1.6 because of sktime
 jinja2>=1.2  # Required by pycaret.internal.utils --> pandas.io.formats.style
 scipy<1.9.0  # Can't >=1.9.0 due to sktime
-joblib>=1.1.0
+joblib>=1.2.0 #The package joblib from 0 and before 1.2.0 is vulnerable to Arbitrary Code Execution (https://github.com/advisories/GHSA-6hrg-qmvc-2xh8)
```
Yard1, line 9:
```suggestion
joblib>=1.2.0  # joblib<1.2.0 is vulnerable to Arbitrary Code Execution (https://github.com/advisories/GHSA-6hrg-qmvc-2xh8)
```

PR #3081, opened 2022-11-10
Fix incompatible types

daikikatsuragawa:
Signed-off-by: Daiki Katsuragawa <50144563+daikikatsuragawa@users.noreply.github.com>

I found some incompatible types when I checked the type hints. This PR is a fix for them.

### DIFF
--- a/pycaret/internal/display/display_backend.py
+++ b/pycaret/internal/display/display_backend.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from pprint import pprint
-from typing import Any, Optional, Union
+from typing import Any, Dict, Optional, Union
 
 import pandas as pd
 
@@ -187,10 +187,6 @@ class DatabricksBackend(JupyterBackend):
         return obj
 
 
-backends = [CLIBackend, JupyterBackend, ColabBackend, DatabricksBackend, SilentBackend]
-backends = {b.id: b for b in backends}
-
-
 def detect_backend(
     backend: Optional[Union[str, DisplayBackend]] = None
 ) -> DisplayBackend:
@@ -209,6 +205,16 @@ def detect_backend(
         return JupyterBackend()
 
     if isinstance(backend, str):
+        backends: Dict[str, Any] = {
+            b.id: b
+            for b in {
+                CLIBackend,
+                JupyterBackend,
+                ColabBackend,
+                DatabricksBackend,
+                SilentBackend,
+            }
+        }
         backend_id = backend.lower()
         backend = backends.get(backend_id, None)
         if not backend:
--- a/pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
+++ b/pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
@@ -119,9 +119,7 @@ class TSForecastingPreprocessor:
         ):
             self._add_transformation_steps(transform=transform_exogenous, target=False)
 
-    def _add_transformation_steps(
-        self, transform: Optional[Union[str, int, float]], target: bool = True
-    ):
+    def _add_transformation_steps(self, transform: str, target: bool = True):
         """Power transform the data to be more Gaussian-like.
 
         Parameters
@@ -189,9 +187,7 @@ class TSForecastingPreprocessor:
         if exogenous_present == TSExogenousPresent.YES and scale_exogenous is not None:
             self._add_scaling_steps(scale=scale_exogenous, target=False)
 
-    def _add_scaling_steps(
-        self, scale: Optional[Union[str, int, float]], target: bool = True
-    ):
+    def _add_scaling_steps(self, scale: str, target: bool = True):
         """Scale the data.
 
         Parameters
--- a/pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
+++ b/pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
@@ -119,7 +119,9 @@ class TSForecastingPreprocessor:
         ):
             self._add_transformation_steps(transform=transform_exogenous, target=False)
 
-    def _add_transformation_steps(self, transform: str, target: bool = True):
+    def _add_transformation_steps(
+        self, transform: Optional[Union[str, int, float]], target: bool = True
+    ):
         """Power transform the data to be more Gaussian-like.
 
         Parameters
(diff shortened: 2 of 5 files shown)


pycaret/internal/display/display_backend.py
```diff
@@ -177,10 +177,6 @@ def _handle_input(self, obj: Any) -> Any:
         return obj
 
 
-backends = [CLIBackend, JupyterBackend, ColabBackend, DatabricksBackend, SilentBackend]
```
daikikatsuragawa, line 180: Example: the type of `backends` was a list or a dictionary.

In addition to that, it does not have to be a global variable because of its usage, so it should be moved.

pycaret/utils/time_series/__init__.py
```diff
@@ -118,7 +118,7 @@ def get_diffs(
 
 def _get_diff_name_list(
     data: pd.Series, data_name: Optional[str] = None, data_kwargs: Optional[Dict] = None
-) -> Tuple[List[pd.Series], List[str]]:
+) -> Tuple[List[pd.Series], List[Optional[str]]]:
```
ngupta23, line 121: I think it will always have data in it. Will it ever be optional?

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
```diff
@@ -119,7 +119,9 @@ def _transformation(
         ):
             self._add_transformation_steps(transform=transform_exogenous, target=False)
 
-    def _add_transformation_steps(self, transform: str, target: bool = True):
+    def _add_transformation_steps(
+        self, transform: Optional[Union[str, int, float]], target: bool = True
```
ngupta23, line 123: I do not think the types are correct here. "transform" has to be a string. Why do you think it should include int and float and why is it Optional?

pycaret/utils/time_series/__init__.py
daikikatsuragawa: @ngupta23 
Thank you for checking!

> I think it will always have data in it. Will it ever be optional?

The type of the argument `data_name` of `_get_diff_name_list` is `Optional[str]`, which defaults to `None`. Therefore, in the following process, the type will be as I have fixed it.

```python
    else:
        # Issue with reconciliation of orders and diffs
        diff_list = [data]
        name_list = [data_name]

    return diff_list, name_list
```

Specifically, the code says that `name_list` may have the value `[None]`.
There may be a mistake in my proposed type or logic. Please check here.

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
daikikatsuragawa: Sorry, the type transform may not have been `Optional`. The process of calling `_add_transformation_steps` is as follows. The value entered here as transform is `transform_target` or `transform_exogenous`. Type `transform_target` and `transform_exogenous` are `Optional[Union[str, int, float]]`. Therefore, I defined the type `Optional[Union[str, int, float]]` for `transform`.

```python
    def _transformation(
        self,
        transform_target: Optional[Union[str, int, float]],
        transform_exogenous: Optional[Union[str, int, float]],
        exogenous_present: bool,
    ):
        # Impute target ----
        if transform_target is not None:
            self._add_transformation_steps(transform=transform_target, target=True)

        # Impute Exogenous ----
        # Only add exogenous pipeline steps if exogenous variables are present.
        if (
            exogenous_present == TSExogenousPresent.YES
            and transform_exogenous is not None
        ):
            self._add_transformation_steps(transform=transform_exogenous, target=False)
```

Also, the following types are being checked in the function named `_add_transformation_steps` and need to be corrected as I suggested. Alternatively, it would be preferable to check the types outside of the function named `_add_transformation_steps`.

https://github.com/pycaret/pycaret/blob/6d1e192103560250862b63f39e065c35b1c8bc2b/pycaret/internal/preprocess/time_series/forecasting/preprocessor.py#L151

ngupta23: > Sorry, the type transform may not have been Optional. The process of calling _add_transformation_steps is as follows. The value entered here as transform is transform_target or transform_exogenous. Type transform_target and transform_exogenous are Optional[Union[str, int, float]]. Therefore, I defined the type Optional[Union[str, int, float]] for transform

But there is a check for None before calling the function so the input to `_add_transformation_steps` should not be Optional.

https://github.com/pycaret/pycaret/blob/a8d7f0def72030e7562898a40bd001768c790a42/pycaret/internal/preprocess/time_series/forecasting/preprocessor.py#L111

> Also, the following types are being checked in the function named _add_transformation_steps and need to be corrected as I suggested. Alternatively, it would be preferable to check the types outside of the function named _add_transformation_steps.

This function is called from the `setup` itself, so it should be fine. From a users flow perspective, they will be notified of the error immediately.

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
```diff
@@ -187,7 +189,9 @@ def _scaling(
         if exogenous_present == TSExogenousPresent.YES and scale_exogenous is not None:
             self._add_scaling_steps(scale=scale_exogenous, target=False)
 
-    def _add_scaling_steps(self, scale: str, target: bool = True):
+    def _add_scaling_steps(
+        self, scale: Optional[Union[str, int, float]], target: bool = True
```
ngupta23, line 193: This should not be int and float and not Optional either since we check for it before passing to this function.

https://github.com/pycaret/pycaret/blob/master/pycaret/time_series/forecasting/oop.py#L1250-L1255

https://github.com/pycaret/pycaret/blob/master/pycaret/internal/preprocess/time_series/forecasting/preprocessor.py#L181-L189

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
daikikatsuragawa: @ngupta23 
Regarding this I have fix based on comments. 

https://github.com/pycaret/pycaret/pull/3081/commits/20dd505f5c6a4be3200526069ceac39f19367991

Thank you.

pycaret/internal/preprocess/time_series/forecasting/preprocessor.py
daikikatsuragawa: @ngupta23 
Regarding this I have fix based on comments. 

https://github.com/pycaret/pycaret/pull/3081/commits/20dd505f5c6a4be3200526069ceac39f19367991

Thank you.

tvdboom (approved):
lgtm

Yard1 (approved):

PR #3172, opened 2022-12-12
Return figure object in case of Streamlit to provide more flexibility

mswastik:
Gives more flexibility on where to place the plot within Streamlit. One can use either `st.plotly_chart(exp.plot_model(*))` or `col.plotly_chart(exp.plot_model(*))` where `col` belongs to instance of `st.columns`

# Related Issue or bug

Info about Issue or bug

Closes #[issue number that will be closed through this PR]

A clear and concise description of what you have done to successfully close your assigned issue. Any new files? or anything you feel to let us know!

# Type of change

<!--

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/pycaret/time_series/forecasting/oop.py
+++ b/pycaret/time_series/forecasting/oop.py
@@ -3583,7 +3583,7 @@ class TSForecastingExperiment(_SupervisedExperiment, TSForecastingPreprocessor):
 
             elif system:
                 if display_format == "streamlit":
-                    st.write(fig)
+                    return fig
                 elif display_format == "plotly-widget":
                     fig.update_layout(autosize=True)
                     ipython_display(


ngupta23:
Hi @mswastik ,

I do not think this is needed. The figure can be returned to you using this argument in plot_model. Let me know if that works.

image

mswastik:
Hi @ngupta23, I tried that argument, the problem is that it leads to generation of two plots in streamlit.

ngupta23:
Can you share your code that generates 2 plots?

mswastik:
`c1,c2=st.columns([2,3])`
`c2.plotly_chart(eda.plot_model(plot="diagnostics", fig_kwargs={'height':900},return_fig=True, display_format='streamlit'), use_container_width=True)`

Without `display_format='streamlit'` the plot displays in new tab, without `return_fig=True` we get error. With both options present we get two plots. 
Probably better way would be to introduce one more condition check that if `return_fig=True` is provided not to execute `st.write` command.

ngupta23:
> Without `display_format='streamlit'` the plot displays in new tab, without `return_fig=True` we get error. With both options present we get two plots.
> Probably better way would be to introduce one more condition check that if `return_fig=True` is provided not to execute `st.write` command.

Hi, Can you share the full reproducible example so I can try this please? Thanks!

ngupta23:
I think the problem is that it is returning the figure as well as plotting it. Maybe returning it should not plot it. I will have a look at it and get back.

moezali1:

PR #3205, opened 2022-12-27
A jarman/issue1766

### LINKED ISSUE #1766: Add lower limit and upper limit for time series forecasting

AJarman:
## Related Issue or bug

> Hi @AJarman
> 
> I discussed this with @moezali1 and @Yard1 in the meeting on 20220409 and this is the recommendation. We should add an argument to setup called `forecast_limit: Optional[List[Union[int, float]] = None`. By default, no limits are applied (same as today's functionality). If user wants to set limits, they can do it as follows:
> 
> ```python
> forecast_limit = None # default - no limits
> forecast_limit = [0, 10000000] # lower and upper limit
> forecast_limit = [0, None]  # lower limit only
> forecast_limit = [None, 10000000] # upper limit only
> ```
> 
> ## Coding Actions:
> * Inside `setup`, there will have to be a check to make sure that if a list is passed, it has exactly 2 entries and the types are correct.
> * If this is enabled, we should add a step to the transformation pipeline. The order of transformations will be as follows
>   
>   * Impute
>   * Limit
>   * Transform
>   * Scale
> 
> ## Future enhancements and user customization
> In the future, we will also extend the following arguments to accept Transformer objects directly so any user wanting to set custom arguments, can pass the object directly (this is similar to what the other modules such as regression and classification do in pycaret). This will do away with the need for the kwargs in setup.
> 
> ```python
> transform_target = "box_cox"  # Today
> transform_target = BoxCoxTransformer(limits = Sth, method="pearsonr")  # In the future (in lieu of transform_kwargs)
> 
> forecast_limit = [0, None] # current proposal
> forecast_limit = ScaledLogitTransformer(0, None) # In the future (in lieu of transform_kwargs)
> ```
> 
> Let me know if this makes sense. Would be happy to discuss more if needed.
> 
> Thanks!

Closes #1766

I have implemented the sktime `scaledLogitTransformer` in the time series module. This is a pipeline step added after imputatation, before transformation and scaling. This currently only affects the target variable and is specified using a list of two values.
Implementing the same for exogenous variables would also be straightforward, however at the moment I'm not sure how this would function in practice, given that they would presumably need different limits.
I have left a 'stubbed' out option (elif) which would be used to implement passing an sktime estimator directly as specified in the future requirements, at present this raises a specific `NotImplementedError`, rather than the generic `TypeError` it would do for any other type.

## Type of change

## How Has This Been Tested?

Unit tests have been added which use the airline dataset from conftest.py as a fixture.
These test a combination of relevant inputs.

NA

## Checklist:

- [] New and existing unit tests pass locally with my changes.

## Screenshots

 Original           | Updated


ngupta23:
Hi Andrew, Thanks for the PR. Could you check on the failing tests please?

moezali1:

PR #3211, opened 2022-12-29
Added DagsHub Logger Support

jinensetpal:

1. Log data and model artifacts to DagsHub Storage.
2. Log experiments to an MLFlow remote hosted by DagsHub.

# Type of change

# Checklist:


Yard1:
Please let me know once this is ready for our review!

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,109 @@
...
+        mlflow.set_tracking_uri(self.remote)
+        super().init_experiment(exp_name_log, full_name)
+
+    def log_artifact(self, file, type="artifact"):
+        def dvc_upload(local_path="", remote_path="", commit=""):
+            assert os.path.isfile(local_path), FileExistsError(
+                f"Invalid file path: {local_path}"
+            )
+            self.dvc_fld.add(file=local_path, path=remote_path)
+            self.dvc_fld.commit(commit, versioning="dvc", force=True)
```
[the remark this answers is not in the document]
timho102003: `force=True` can avoid the additional check for whether the uploaded file is the same as the one on the remote server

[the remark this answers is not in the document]
nirbarazida: Why would we want to avoid it? If the file is already pushed, I don't see a reason to override it, and if it's a new version, then there is no need fo force.

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,108 @@
...
+            and os.getenv("MLFLOW_TRACKING_PASSWORD") is not None
+        )
+        if not is_mlflow_set or remote is None:
+            prompt_in = input("Please insert your repository owner_name/repo_name:")
+            prompt_in = prompt_in.split("/")
+            assert (
+                len(prompt_in) == 2
+            ), f"Invalid input, should be owner_name/repo_name, but get {prompt_in} instead"
+            os.environ["REPO_OWNER"] = prompt_in[0]
+            os.environ["REPO_NAME"] = prompt_in[1]
```
nirbarazida: Why are we saving those variables as environment variables?

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,102 @@
...
+        super().__init__()
+
+        # check token exist or not:
+        from dagshub.auth.tokens import TokenStorage, _get_token_storage
+
+        token_dict = _get_token_storage()._load_cache_file()
+        is_token_set = not TokenStorage._is_expired(token_dict)  # Check OAuth
+        if not is_token_set:
+            dagshub.auth.get_token()
+
```
nirbarazida, line 28: These are privet functions, instead we should use the dagshub.common.config to check if the user is logged in.

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,102 @@
...
+            
+            dagshub.init(repo_name=prompt_in[0], repo_owner=prompt_in[1])
+            remote = os.getenv("MLFLOW_TRACKING_URI")
+
+        from dagshub.upload import Repo
+
+        self.run = None
+
+        self.remote = remote
+        self.__dvc_fld_path = Path("artifacts")
```
nirbarazida, line 47: What does `fld` mean? The variables should be more descriptive

nirbarazida (changes requested):

pycaret/internal/pycaret_experiment/tabular_experiment.py
```diff
@@ -2389,14 +2393,18 @@ def save_model(
...
             model_name=model_name,
             prep_pipe_=None if model_only else pipeline_to_use,
             verbose=verbose,
             use_case=self._ml_usecase,
             **kwargs,
         )
+        if self.logging_param:
+            self.logging_param.log_remote_artifact(model_name, type="model")
+
+        return model_, model_name
```
Yard1, line 2407:
```suggestion
        model_, model_filename = pycaret.internal.persistence.save_model(
            model=model,
            model_name=model_name,
            prep_pipe_=None if model_only else pipeline_to_use,
            verbose=verbose,
            use_case=self._ml_usecase,
            **kwargs,
        )
        if self.logging_param:
            self.logging_param.log_remote_artifact(model_filename, type="model")

        return model_, model_filename
```

pycaret/loggers/dashboard_logger.py
```diff
@@ -257,11 +263,34 @@ def log_experiment(
...
 
+                    # upload data to remote server
+                    self.log_remote_artifact(test_path, type="data")
+                    if experiment.transform_target_param:
+                        train_transform_path = os.path.join(
+                            tmpdir, "Dataset_transform.csv"
+                        )
+                        experiment.train_transformed.to_csv(train_transform_path)
+                        self.log_remote_artifact(train_transform_path, type="data")
+                self.log_remote_artifact(filename="", type="data_commit")
```
Yard1, line 292: what is the purpose of this call?

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,94 @@
...
+class DagshubLogger(MlflowLogger):
+    def __init__(self, remote=None) -> None:
+        super().__init__()
+        if dagshub is None:
+            raise ImportError(
+                "DagshubLogger requires dagshub. Install using `pip install dagshub`"
+            )
+        # check token exist or not:
+        token = dagshub.auth.get_token()
+        os.environ["MLFLOW_TRACKING_USERNAME"] = token
```
Yard1, line 23: is it possible to avoid modifying env vars?

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,94 @@
...
+        if not remote or "dagshub" not in os.getenv("MLFLOW_TRACKING_URI"):
+            prompt_in = input(
+                "Please insert your repository owner_name/repo_name:"
+            ).split("/")
+            assert (
+                len(prompt_in) == 2
+            ), f"Invalid input, should be owner_name/repo_name, but get {prompt_in} instead"
+
+            dagshub.init(repo_name=prompt_in[1], repo_owner=prompt_in[0])
+            remote = os.getenv("MLFLOW_TRACKING_URI")
```
Yard1, line 36: Can we add a path to do this programatically? eg. passing `owner_name/repo_name` as a string argument.

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,94 @@
...
+try:
+    import dagshub
+    import mlflow
+    from dagshub.upload import Repo
+except ImportError:
+    dagshub = None
+
+
+class DagshubLogger(MlflowLogger):
+    def __init__(self, remote=None) -> None:
```
Yard1, line 15: In general, to avoid serialization issues, we should keep any actual initialisation logic in `init_experiment`. Ideally, `__init__` would just set attributes.

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,94 @@
...
+        self.__commit_data_type = []
+
+    def init_experiment(self, *args, **kargs):
+        mlflow.set_tracking_uri(self.remote)
+        super().init_experiment(*args, **kargs)
+
+    def _dvc_add(self, local_path="", remote_path=""):
+        assert os.path.isfile(local_path), FileExistsError(
+            f"Invalid file path: {local_path}"
+        )
```
Yard1, line 62: can we use normal `raise` syntax here?

pycaret/loggers/dagshub_logger.py
```diff
@@ -0,0 +1,94 @@
...
+        # Check mlflow environment variable is set:
+        if not remote or "dagshub" not in os.getenv("MLFLOW_TRACKING_URI"):
+            prompt_in = input(
+                "Please insert your repository owner_name/repo_name:"
+            ).split("/")
+            assert (
+                len(prompt_in) == 2
+            ), f"Invalid input, should be owner_name/repo_name, but get {prompt_in} instead"
+
+            dagshub.init(repo_name=prompt_in[1], repo_owner=prompt_in[0])
```
Yard1, line 35: what happens if repo already exists?

Yard1 (commented):
Thanks, this looks good! Left some comments.

pycaret/loggers/dagshub_logger.py
jinensetpal: We could try caching or writing to temporary storage, but I'd assume that'll make the code needlessly convoluted

pycaret/loggers/dagshub_logger.py
jinensetpal: Will fix!

pycaret/loggers/dagshub_logger.py
jinensetpal: Will fix!

pycaret/loggers/dashboard_logger.py
jinensetpal: @timho102003

pycaret/loggers/dagshub_logger.py
jinensetpal: `dagshub.init` manages this. The repository is created only if it does not exist. If it does exist, the environment variables are set up per usual and they can continue to log additional experiments and artifacts to that repository.

pycaret/loggers/dagshub_logger.py
jinensetpal: We did discuss this internally! We noted that it would require us to update the `setup` function - but this variable would only be used if we were logging using DagsHub, so it felt a bit unwieldy. WDYT?

Yard1: It is possible to pass the logger object directly as well, so there's that escape hatch. I don't think we need to modify the `setup` function here - I am fine with `input` being the primary mode of operation, but it shouldn't be the only one :)

pycaret/loggers/dagshub_logger.py
nirbarazida: MLflow uses env vars to log experiments to remote servers. It's the default behavior defined by MLflow. I think we shouldn't reinvent the wheel and stick with the flow MLflow users are used to.

pycaret/loggers/dagshub_logger.py
nirbarazida: @jinensetpal let's add it as a var for the DagsHub logger object with default value = None.
If it's None, then ask for input.

jinensetpal: Perfect

Yard1 (approved):
Thanks, LGTM! Would still like a reply to this https://github.com/pycaret/pycaret/pull/3211#discussion_r1062357555 though :)

Yard1:
@jinensetpal @timho102003 I am happy to merge this as-is, but in a followup we should try to move the `log_remote_artifact` logic to `DagshubLogger` itself instead of having it in `DashboardLogger`. My reasoning here is that this is, so far, unique to DagsHub and the remote part of the logic could instead be put in `DagshubLogger.log_artifact`.

Looking at the code, there is 1:1 equivalence between `log_artifact` and `log_remote_artifact` calls, aside from the latter being applied to transformed data as well - we can log that with `log_artifact` as well.

As an unrelated aside, we should make logging more configurable - for example, datasets may be quite big and users may not want to log them (this should be done in a separate PR).

timho102003:
I agree with moving `log_remote_artifact` to `log_artifact `
For the dataset part, the users can choose to set `log_data=False` in the experiment `setup()` if they don't want to log them

PR #3214, opened 2022-12-29
initial implementation for #3202

### LINKED ISSUE #3202: [ENH]: Prevent possible leakage in Time Series analysis based on use case

Currently during the setup process, we use the entire y dataset (train + test) to determine some high level model parameters. e.g. seasonal periods, d, D, etc.

https://github.com/pycaret/pycaret/blob/edabf1a2e1779fc83e70946e367ec82f69a9a3f6/pycaret/time_series/forecasting/oop.py#L1929-L1931

### For seasonal period detection:

https://github.com/pycaret/pycaret/blob/edabf1a2e1779fc83e70946e367ec82f69a9a3f6/pycaret/time_series/forecasting/oop.py#L703

https://github.com/pycaret/pycaret/blob/edabf1a2e1779fc83e70946e367ec82f69a9a3f6/pycaret/time_series/forecasting/oop.py#L723-L726

### For finding multiplicative components

https://github.com/pycaret/pycaret/blob/edabf1a2e1779fc83e70946e367ec82f69a9a3f6/pycaret/time_series/forecasting/oop.py#L1139

### Setup EDA

https://github.com/pycaret/pycaret/blob/edabf1a2e1779fc83e70946e367ec82f69a9a3f6/pycaret/time_series/forecasting/oop.py#L1217-L1229

These hyperparameters like `sp` have to be determined ahead of time since finding them using brute force or grid search would take a long time. We also do not want these "hyperparameters" to change from one CV fold another or from train to test split since the underlying process generating the data should remain constant no matter which time period we are looking at.

### Use Case 1: When doing the forecast after finalizing the model.

Based on the current approach of using the entire dataset for determining these "hyperparameters", the test and CV metrics may be a little optimistic (since entire test and train data was used to calculate these hyperparameters), but the goal is to get good future unknown forecasts so it "may" be best from that point of view. There is definitely a tradeoff here.

### Use Case 2: Benchmarking

In this case, the user does not want to use a test set or do cross validation. They want to use the entire data, fit the model(s) of choice and get the predictions without finalizing the model - i.e. in this case, the model from create_model would be considered the finalized model. This is ideal when you want to get the default results from pycaret in the fastest way possible (possibly for benchmarking). e.g. below

```python
exp = TSForecastingExperiment()
exp.setup(data, fh=12, session_id=42)
model = exp.create_model("ets")
preds = exp.predict_model()
```

That is it. No finalizing, just get the results in the fastest possible way. If I were to go through the regular methodology of doing CV folds, tune_model and then finalize_model, it would add extra compute time that will look bad from a benchmarking perspective. Ideally, the default model will be robust enough to give the best (or close to best) predictions as well since that is also important from a benchmarking point of view. 

This is currently not possible in pycaret since we split the entire dataset into train and test by default and test set can not be 0. We could possibly provide the test set time points only, with NA values for y, but then this would require imputation of test set (since that is a pre requisite of pycaret time series). Then the hyperparameter estimation will be impacted by the quality of the imputed test set points. This is undesirable. 

#### Possible alternate 1

We can change the calculation of these hyperparameters (e.g. sp) to use only train dataset. In this case, the test metrics will be unbiased, but the CV metrics will be biased (since we have used all the train data to determine these hyperparameters).

#### Possible alternate 2

We can change the calculation of these hyperparameters (e.g. sp) to use only the train part of the first CV fold in the data. This would result in truly unbiased CV and test metrics. But it would severely hamper the quality of these hyperparameters since we have looking at possibly a very small dataset for calculating them. This is likely to impact the forecasts after finalizing the model and that is what we care about eventually.

### My Recommendation

To resolve these use cases and provide flexibility to the users, we should provide them with all three options.

1. **Solution 1: Estimate the hyperparameters using the entire dataset**. This is the current implementation and uses a tradeoff between getting best unknown future predictions and getting slightly biased metrics.

2. **Solution 2: Estimate the hyperparameters using only the train split.** This will help solve Use Case 2 (benchmarking). In this case, the user can pass only test time points without y_test values. Even though these will need to be imputed, we will not use these imputed test values to estimate the hyperparameters.

3. **Solution 3: Estimate the hyperparameters using only the train part of the first CV split.** This will help when the user has a lot of data and can forgo data after the train part of the first CV split.

_No response_

### Additional context

_No response_

ngupta23:
# Related Issue or bug

Info about Issue or bug

Closes #3202

Added ability to select the split to be used to determine the hyperparameters

# Type of change

# How Has This Been Tested?

Unit test have been added and existing tests need to be updated.

# Checklist:

### DIFF
--- a/pycaret/internal/tests/time_series.py
+++ b/pycaret/internal/tests/time_series.py
@@ -1,5 +1,6 @@
 from typing import Dict, List, Optional, Tuple, Union
 
+import numpy as np
 import pandas as pd
 import statsmodels.api as sm
 from pmdarima.arima.utils import ndiffs, nsdiffs
@@ -463,7 +464,7 @@ def _is_white_noise(
     data : pd.Series
         Time Series data on which the test needs to be performed
     lags : List[int], optional
-        The lags used to test the autocorelation for white noise, by default [24, 48]
+        The lags used to test the autocorrelation for white noise, by default [24, 48]
     alpha : float, optional
         Significance Level, by default 0.05
     verbose : bool, optional
@@ -518,7 +519,10 @@ def _is_white_noise(
             )
 
         # Step 2C: Cleanup results ----
-        results[test_category] = results["lb_pvalue"] > alpha
+        # When all values are same, results are NA, hence replace with True
+        results[test_category] = np.where(
+            ~results["lb_pvalue"].isna(), results["lb_pvalue"] > alpha, True
+        )
         is_white_noise = False if results[test_category].all() is False else True
         results.rename(
             columns={"lb_stat": "Test Statictic", "lb_pvalue": "p-value"},
--- a/pycaret/time_series/forecasting/functional.py
+++ b/pycaret/time_series/forecasting/functional.py
@@ -52,6 +52,7 @@ def setup(
     fold_strategy: Union[str, Any] = "expanding",
     fold: int = 3,
     fh: Optional[Union[List[int], int, np.ndarray, "ForecastingHorizon"]] = 1,
+    hyperparameter_split: str = "all",
     seasonal_period: Optional[Union[List[Union[int, str]], int, str]] = None,
     sp_detection: str = "auto",
     max_sp_to_consider: Optional[int] = None,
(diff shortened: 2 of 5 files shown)


Yard1 (approved):
Looks good to me, thanks!

PR #3238, opened 2023-01-09
Less restrictive column character validation

tvdboom:
# Related Issue or bug

Setup and create_model could fail for datasets with weird chars in the column names. The new restriction on chars is more flexible. The only (prohibited) chars are those that make lightgbm fail: `[]{}:",`


pycaret/internal/preprocess/transformers.py
```diff
@@ -266,7 +266,7 @@ def inverse_transform(self, y):
 class CleanColumnNames(BaseEstimator, TransformerMixin):
     """Remove weird characters from column names."""
 
-    def __init__(self, match="[^A-Za-z0-9_]+"):
+    def __init__(self, match=r"[\]\[\,\{\}"\:]+"):
```
Yard1, line 269:
```suggestion
    def __init__(self, match=r"[\]\[\,\{\}\"\:]+"):
```

PR #3252, opened 2023-01-13
added notebook tests and pre-commit-hooks for isort and flake8

tvdboom:

* Added pre-commit-hooks for isort and flake8
* Added a test to CI to check that the notebooks in the examples dir run successfully.

**Important**
This PR requires future PRs to have all example notebooks correctly running. This means that @moezali1 we should clean the examples dir from all old/unused notebooks and update the rest to make them work with `master`.

Also, it's possible to create a pre-commit-hook to run the notebooks and check them (even persist the results), but I think this isn't desirable since the notebooks can run for a long time and it would make commits take forever.

# Type of change


moezali1:
@tvdboom @ngupta23 @Yard1 What do you guys think about this:

- just like we have created a separate [repo](https://github.com/pycaret/datasets) for datasets. I propose we should create separate repo for examples and call it `pycaret-examples`. We will move out all the notebooks in that repo.

The only notebook that will stay in pycaret repo will be official tutorials (english only) and they will be part of unit test. What do u guys think?

ngupta23:
> The only notebook that will stay in pycaret repo will be official tutorials (english only) and they will be part of unit test. What do u guys think?

- The 20x notebooks in time series still take time to run due to large dataset. Will need to be reworked with a simpler dataset to reduce run time


> just like we have created a separate [repo](https://github.com/pycaret/datasets) for datasets. I propose we should create separate repo for examples and call it pycaret-examples. We will move out all the notebooks in that repo.

- Will the have a separate job running in that repo for this to check for whether these continue to run or not (in case we make changes to the pycaret library in the future)?

.github/workflows/test.yml
```diff
@@ -37,6 +37,22 @@ jobs:
...
+      - name: Set up Python environment
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: |
+          pip install -U pytest pytest-xdist nbmake
+          pip install -e ".[full]"
+      - name: Run notebooks
+        run: pytest --nbmake -n=auto --ignore=./tutorials/japanese/ --ignore=./tutorials/chinese/ --ignore=./tutorials/greek/ --ignore=./tutorials/portuguese/ ./tutorials/
```
tvdboom, line 54: @Yard1 @ngupta23 Any idea how to exclude subfolders with one ignore or directly in the path? with some wildcard magic perhaps...

ngupta-o9: Can you delete the subfolders in the installation steps (like we do for some of the unit tests)?

moezali1 (approved):

PR #3350, opened 2023-02-28
add missing package to requirements. Fixes #3349

### LINKED ISSUE #3349: [INSTALL]: Test discovery fails due to a missing package.
### Installation check

- [X] I have read the [installation guide](https://pycaret.gitbook.io/docs/get-started/installation).

### Platform

Windows-10-10.0.19045-SP0

### Installation Method

pip install

### pycaret Version

masterr branch / '3.0.0.rc9'

### Python Version

Python 3.10.9

### Description

Setup of the contribution environment fails.

Test discovery fails due to a missing package.

The package is not listed in the `requirements-test.txt` file.

Missing package is `daal4py`.

### Installation Logs

<details>

=================================== ERRORS ====================================
_____________ ERROR collecting tests/test_time_series_engines.py ______________
ImportError while importing test module 'e:\dev\pycaret\tests\test_time_series_engines.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
C:\Users\Pascal\AppData\Local\Programs\Python\Python310\lib\importlib\__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests\test_time_series_engines.py:3: in <module>
    from daal4py.sklearn.linear_model._linear import (
E   ModuleNotFoundError: No module named 'daal4py.sklearn.linear_model._linear'
============================== warnings summary ===============================

</details>

pascal456:
# Related Issue or bug

Test discovery fails due to missing package.

Closes #3349

Resolved by adding the missing package `daal4py`.

# Type of change

# How Has This Been Tested?

After the change I was able to install with `pip install -e .[test]` and run the test discovery successfully.

`NA`

# Checklist:

# Screenshots

 Original           | Updated


pascal456:
The changes in `requirements-test.txt` look a little cluttered because I also sorted the lines in ascending alphabetical order.

The actual line added is
```
daal4py==2023.0.1
```

requirements-test.txt
```diff
@@ -1,5 +1,6 @@
-pytest>=7.0.1
-moto>=3.0.7
 codecov>=2.1.12
+daal4py==2023.0.1
```
ngupta23, line 2: These are already part of requirements-optional.txt. You should install it either manually in your local environment (my preferred approach) or install the entire requirements-optional.txt file.

Also note that MLFlow is another optional library that causes test discovery to fail. This is also covered in requirements-optional.txt hence, it is not needed here.

ngupta23:
@pascal456 Thanks for your first contribution to the repository!

I left a comment on the change. I don't think we should change the requirements-test.txt file. Instead, if you want, we should update the test to only run if the required optional libraries are installed. One place where this may have been done is linked below, but I have not explored it in detail. If you wish, you can evaluate these and propose a change to the tests to handle these optional dependencies. Any other approach to handle the soft dependencies (ignore tests if not installed) is also fine. You can make a proposal if you wish.

Thanks!

https://github.com/sktime/sktime/blob/main/sktime/tests/test_softdeps.py

Yard1:
Fully agree with @ngupta23 that the best solution here would be to make pytest skip the tests if the package cannot be imported. Something like this could also be considered (probably the easiest solution) - https://www.scivision.dev/pytest-skip-importerror/

pascal456:
OK. But I have another idea, because the problem originated from the documentation, that I followed. If you want to avoid adding it, then the contributing documentation should be changed. Maybe this should be enough then, and also more appropriate. Others will face this issue as well.

Skipping tests, because modules are not installed is like ignoring a warning sign in traffic becuase I do not understand the language it is written in. The problems they are handling, are still valid and could still appear.

it is [documented to install the package in the following way](https://github.com/pycaret/pycaret/blob/master/CONTRIBUTING.md):
```sh
pip install -e .[test]
```

And this did not work for me. Therefore I suggested to add it to the `requirements-test.txt`. So the correct installation call mentioned there should be 
```sh
pip install -e .[dev]
```

requirements-test.txt
pascal456: just checked. daal4py is **not** in the `requirements-optional.txt` file. Or is it a dependency of another package?

ngupta23:
> Skipping tests, because modules are not installed is like ignoring a warning sign in traffic becuase I do not understand the language it is written in. The problems they are handling, are still valid and could still appear.

Agreed. The unit tests on GitHub CI do not ignore any tests since they install all libraries including optional ones. This recommendation is only to avoid test discovery locally on IDEs like VSCode due to optional dependencies not being installed locally.

requirements-test.txt
ngupta23: I think it is installed by this optional dependency

```
scikit-learn-intelex
```

PR #3465, opened 2023-04-07
Restrict evidently version

mike0sv:
In https://github.com/evidentlyai/evidently/pull/550 old API will be removed, so you need to restrict evidently version before you switch to new API


Yard1 (approved):

Yard1:
Thanks!

Issue #3433, opened 2023-03-26
[BUG]: Broken pipeline in catboost regressor


aquinteros:
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).

- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.

- [X] I have confirmed this bug exists on the master branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@master).

### Issue Description

Hello! i checked and didn't find the issue reported before, i'm very sorry if i couldn't find it.

Also i apologize for my broken english.

The pipeline i am working on ran perfecly with last version of pycaret, but when i updated my model to pycaret 3.0.0 it stopped working.

To make it reproducible i will list the basic tasks i did:

I started a new venv in python 3.10.10 locally in a folder and installed the following list of requirements.txt in this order:
pandas
numpy
sqlalchemy
scikit-learn
seaborn
matplotlib
datetime
streamlit
pyodbc
sqlalchemy<2.0
mlflow
xgboost
pycaret[full]

When i ran my pipeline it selected catboost regressor as the estimator by r2, i tuned the model, finalized it and saved it to a subfolder "models"

When i tried to load the model using load_model() it loadad correctly but when i ran the predict_model() function i got this error:

"ValueError: If estimator is not a Pipeline, you must run setup() first."

i ran the same steps above but in stead of using compare_models i used create_model("lightgbm") and it worked just fine with that algorithm which makes me think its related only to catboost regressor.

### Reproducible Example

```python
# importar librerias
import pandas as pd
import seaborn as sns
import matplotlib.pyplot as plt
import numpy as np

from pycaret.regression import load_model, predict_model
from pycaret.regression import setup, compare_models, tune_model, plot_model, finalize_model, save_model, create_model, evaluate_model

# setup model 
session = setup(
	data = train,
	target = 'Monto_EUR',
	log_experiment = True,
    use_gpu=False,
    session_id=seed,
    normalize = True,
    normalize_method = 'zscore',
    )

# compare models
model = compare_models(exclude=['dummy','ada','en','lar','llar','lasso','rf','et'], sort='r2')

# tune model
model = tune_model(model, optimize='mape', n_iter=100, choose_better=True)

# finalize_model
model = finalize_model(model)

# save_model
save_model(model=model, model_name='models/predictor_model')

# load model
model = load_model('models/predictor_model')

[...]

PR #3481, opened 2023-04-16
Fix `save_model` with catboost

Yard1:
# Related Issue or bug

Info about Issue or bug

Closes https://github.com/pycaret/pycaret/issues/3433

Removes unnecessary `deepcopy` in `save_model` which for some reason have caused issues with catboost. There should be no downside to this change.

# Type of change

# How Has This Been Tested?

# Checklist:

# Screenshots

 Original           | Updated

### DIFF
--- a/pycaret/internal/persistence.py
+++ b/pycaret/internal/persistence.py
@@ -297,39 +297,41 @@ def save_model(
     logger.info("Adding model into prep_pipe")
 
     if use_case == MLUsecase.TIME_SERIES:
-        model_ = deepcopy(model)
         if prep_pipe_:
             pipeline = deepcopy(prep_pipe_)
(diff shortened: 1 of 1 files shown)


pycaret/internal/persistence.py
```diff
@@ -297,39 +297,41 @@ def save_model(
     logger.info("Adding model into prep_pipe")
 
     if use_case == MLUsecase.TIME_SERIES:
-        model_ = deepcopy(model)
```
ngupta23, line 300: Without the copy, will it not override the original model object when we add the pipeline?

Yard1: I think it should be fine, because we don't modify the model object itself (it will just be referenced in multiple pipelines). I will add a test to make sure, though.

Issue #3669, opened 2023-07-27
[BUG]: Time Series plot still being shown when `return_fig=True`


chrimaho:
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).
- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.
- [X] I have confirmed this bug exists on the master branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@master).

### Issue Description

I am trying to plot my time series model.

After I have built the model, I then try to run:

```py
fig = exp.plot_model(
    plot="decomp",
    return_fig=True,
)
fig.show()
```

The result is that the plot will be printed _twice_. This is not desired behaviour.

### Reproducible Example

```python
from pycaret.datasets import get_data
from pycaret.time_series import TSForecastingExperiment
y = get_data('airline')
eda = TSForecastingExperiment()
eda.setup(data=y, fh=12, fold=3)
fig = eda.plot_model(plot='ts', return_fig=True)
fig.show()
```

### Expected Behavior

When `return_fig=True` then nothing will be displayed on the canvas. Because the fig will be assigned as a variable, which will later be displayed with `fig.show()`.

The issue occurs here:

https://github.com/pycaret/pycaret/blob/66621009d3baaf7ca41cd7cd92f4d406481591d7/pycaret/time_series/forecasting/oop.py#L3893

Instead of having this line, it should actually be like this:

```py
if not return_fig:
    fig.show(renderer=renderer)
    self.logger.info("Visual Rendered Successfully")
else:
    self.logger.info("Visual not rendered, but will be returned from function")
```

That way, it will resolve the 'double printed' issue.

### Actual Results

```python-traceback
No error is thrown. This is a functional (logical) bug.
```

### Installed Versions

<details>

```text
System:
    python: 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0]
executable: /mnt/c/Users/user/.../time-series-exploration/.venv/bin/python
   machine: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

PyCaret required dependencies:
                 pip: 23.1.2
          setuptools: 68.0.0
             pycaret: 3.0.4
             IPython: 8.14.0
          ipywidgets: 8.0.7
                tqdm: 4.65.0
               numpy: 1.23.5
              pandas: 1.5.3
[...]

PR #3670, opened 2023-07-27
Resolve double-printed plots during ts experiment

chrimaho:
Closes: #3669

## Changes

Enabled a safe `if` check when running `fig.show()` during TimeSeries experiments. Without this change, then plots will be printed twice when we set the parameter: `return_fig=True`.

Full description included in:
- #3669

## Type of change

## How Has This Been Tested?

No. It wasn't a bug previously, only a logical issue.

## Describe if there is any unusual behaviour of your code

NA

## Checklist:

### DIFF
--- a/pycaret/time_series/forecasting/oop.py
+++ b/pycaret/time_series/forecasting/oop.py
@@ -3894,7 +3894,9 @@ class TSForecastingExperiment(_TSSupervisedExperiment, TSForecastingPreprocessor
                             fig.show(renderer=renderer)
                             self.logger.info("Visual Rendered Successfully")
                         else:
-                            self.logger.info("Visual not rendered, but will be returned from function")
+                            self.logger.info(
+                                "Visual not rendered, but will be returned from function"
+                            )
                     except ValueError as exception:
                         self.logger.info(exception)
                         self.logger.info("Visual Rendered Unsuccessfully")
--- a/pycaret/time_series/forecasting/oop.py
+++ b/pycaret/time_series/forecasting/oop.py
@@ -4006,8 +4006,9 @@ class TSForecastingExperiment(_TSSupervisedExperiment, TSForecastingPreprocessor
             * 'residuals' - Residuals Plot
 
 
-        return_fig: : bool, default = False
+        return_fig: bool, default = False
             When set to True, it returns the figure used for plotting.
+            When set to False (the default), it will print the plot, but not return it.
 
 
         return_data: bool, default = False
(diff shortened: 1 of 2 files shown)


ngupta23:
@chrimaho Could you please fix the Code Quality issues? Thanks!

chrimaho:
> @chrimaho Could you please fix the Code Quality issues? Thanks!

Done 👍  
Thanks!

ngupta23:
I think this change should apply to all options under the `elif system:` block. I remember that [mswastik](https://github.com/mswastik) also reported this earlier but for streamlit (see https://github.com/pycaret/pycaret/pull/3172).

So if `return_fig` is set to True, none of the options should display the figure to the user automatically. The responsibility of displaying will then be on the user, who must catch the returned object and (optionally manipulate and) display it as needed. 

What do you think?

If we do make this change, please make sure to modify the docstring in noth this file (`oop.py`) and `functional.py` to reflect the same.

chrimaho:
Hi @ngupta23 ,
Agreed on all points 👍 
Leave it with me, I'll update it again over the next few days and will ping you when it's ready for you to review again.
Thanks.

chrimaho:
Hi @ngupta23 ,
All updated, ready for your review now.
Thanks.

pycaret/time_series/forecasting/oop.py
```diff
@@ -3857,23 +3857,26 @@ def _plot_model(
 
             elif system:
                 if display_format == "streamlit":
-                    st.write(fig)
+                    if not return_fig:
```
ngupta23, line 3860: Since you are adding this if statement (`if not return_fig`) to every if-else block under (`elif system`), would it make sense to combine it with `elif system`?

e.g. `elif system and not return_fig:`

chrimaho:
Yes, I agree with you.

I hadn't done this initially as I saw that there was additional logic applied in each option within the `elif system:` block. And I wanted to maintain these.

However, thinking this through, all of those logics (eg. setting the `renderer` based on `big_data_threshold`) is done to prevent slowing down the notebook during rendering. And, realistically, when we are returning the `fig` object, we would then want the user to handle this issue, not `pycaret`.

I also saw that in the other Supervised Experiments within `pycaret`, they had this secondary `if` check within their `elif system:` block. For example:

https://github.com/pycaret/pycaret/blob/4697e3874eb5558f51ba9c91cdf57f8658207149/pycaret/internal/pycaret_experiment/tabular_experiment.py#L692-L705

Therefore, I wanted to follow the same. However, as mentioned, these additional logics are not necessary when the `fig` object is being returned. So I will update it to meet your suggestion.

pycaret/time_series/forecasting/oop.py
chrimaho: Done. Now ready for review @ngupta23

ngupta23 (approved):

ngupta23:
Thanks for making the changes. I have merged it with the master branch.

PR #3745, opened 2023-09-22
Fix MLflowClient import

celestinoxp:
Replaced the incorrect import 'from mlflow.tracking.client import MlflowClient' with the correct import 'from mlflow.tracking import MlflowClient'

# Related Issue or bug

This fix should correct the mlflow issue on the test system.


Yard1 (approved):
Thanks!

Issue #3746, opened 2023-09-22
[BUG]: plot_kwargs not passed to yellowbrick visualizers in function plot_model()


morenobonaventura:
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).

- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.

- [X] I have confirmed this bug exists on the master branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@master).

### Issue Description

plot_model() accept plot_kwargs which are passed down to a number of yellowbrick plotting function: KElbowVisualizer, ROCAUC, DiscriminationThreshold etc..
see:
https://github.com/pycaret/pycaret/blob/master/pycaret/internal/pycaret_experiment/tabular_experiment.py#L1049
https://github.com/pycaret/pycaret/blob/master/pycaret/internal/pycaret_experiment/tabular_experiment.py#L1156
https://github.com/pycaret/pycaret/blob/master/pycaret/internal/pycaret_experiment/tabular_experiment.py#L1175

However those kwargs are not passed to the `ValidationCurve()` call
https://github.com/pycaret/pycaret/blob/master/pycaret/internal/pycaret_experiment/tabular_experiment.py#L1748

### Reproducible Example

```python
import numpy as np
from matplotlib import pyplot as plt
from yellowbrick.model_selection import ValidationCurve
from pycaret.classification import *
from pycaret.datasets import get_data


def main():
    # load dataset
    diabetes = get_data('diabetes')

    s = setup(data=diabetes, target='Class variable', session_id=123)

    model = create_model('lightgbm')

    print('plotting yellowbrick learning curve')
    params = {
        'param_name': 'learning_rate',
        'param_range': np.arange(0.1, 1.1, 0.1),
        'scoring': 'precision'
    }
    visualizer = ValidationCurve(model, **params)
    X_train = get_config(variable="X_train")
    y_train = get_config(variable="y_train")
    visualizer.fit(X_train, y_train)
    visualizer.show(outpath='yellowbrick_validation_curve.png')

    plt.clf()
    print('plotting pycaret learning curve')
    plot_model(model, plot='vc', save=True, plot_kwargs=params)


if __name__ == '__main__':
    main()
```

### Expected Behavior

The Validation Curve plot created by pycaret should be equal to the one created with yellowbricks
[...]

PR #3748, opened 2023-09-22
Fix python version between 3.8 and 3.10

celestinoxp:
# Related Issue or bug

This change is to avoid problems with versions of Python that are no longer supported such as Python 3.7


celestinoxp:
example to avoid: https://github.com/pycaret/pycaret/issues/3746
user have python 3.7 and pycaret 3,x... should use python 3.8 or higher.. max 3.10...

tvdboom:
The changes in this PR are also in #3749. You probably forgot to create the new branch from master instead of the other current branch. I'll cancel this PR and let's focus on the other one.

PR #3756, opened 2023-09-28
add python 3.11 support

tvdboom:
# Describe the changes you've made

Add support for python 3.11.

Additionnaly, fix a bug in `test_classification_custom_metric`.

# Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, local variables)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #3813, opened 2023-11-05
Dependencies: Permit NumPy up to 1.26. CI: Run Python 3.11 on `test_linux` workflow.

amotl:
Hi again,

in order to support GH-3756, this patch demonstrates that running the test suite succeeds on Python 3.11 without much ado, by permitting an installation of a more recent version of NumPy. This will probably also open the door for eventually thinking about upgrading pandas.

We exercised it at [^1] already, where Python 3.11/Linux succeeded, and we think the test suite failing on Windows 3.8 might be unrelated. Maybe you can have a closer look?

image

With kind regards,
Andreas.

[^1]: https://github.com/crate-workbench/pycaret/actions/runs/6762683108

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 ipython>=5.5.0
 ipywidgets>=7.6.5  # required by pycaret.internal.display
 tqdm>=4.62.0  # required by pycaret.internal.display
-numpy>=1.21, <1.24  # Can't >=1.24 because of np.float deprecation
+numpy>=1.21, <1.27
 pandas>=1.3.0, <2.0.0  # Can't >=2.0.0 because of sktime
 jinja2>=1.2  # Required by pycaret.internal.utils --> pandas.io.formats.style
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@
 ipython>=5.5.0
 ipywidgets>=7.6.5  # required by pycaret.internal.display
 tqdm>=4.62.0  # required by pycaret.internal.display
-numpy>=1.21, <1.24  # Can't >=1.24 because of np.float deprecation
+numpy>=1.21, <1.27
```
amotl, line 5: We can't see where the code base uses `np.float`. Only `np.float32` and `np.float64` have been discovered so far. So, we think it will be good to go.

Yard1:
Hi, thanks for the PR! Let's see if the CI passes.

amotl:
It looks good so far, but test_windows (3.10) apparently failed again after 14 minutes spending within the "Post Run" actions cleanup task.

Yard1 (approved):

Yard1:
Thanks!

PR #3821, opened 2023-11-15
Add one-click deploy to readme

jbgh:
Hey!

We built this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:

https://13b7fadfffe31ff3c13c01566691ec4a78f664be.dome.tools

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.


moezali1:

PR #3830, opened 2023-11-27
Support Pandas 2.0

### LINKED ISSUE #3722: [BUG]: Support Pandas v2.1
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).

- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.

- [X] I have confirmed this bug exists on the master branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@master).

### Issue Description

sktime is working to support new pandas 2.1, we should do the same (https://github.com/sktime/sktime/pull/5185)

more context: since pandas 1.5.3, have various releases we not support, ex: pandas 2, 2,0.1, 2.0,2, 2.0.3 and now 2.1

time to update!

### Reproducible Example

```python
.
```

### Expected Behavior

.

### Actual Results

```python-traceback
.
```

### Installed Versions

<details>
latest
</details>

celestinoxp:
Closes #3722


pycaret/internal/preprocess/transformers.py
```diff
@@ -585,7 +585,11 @@ def transform(self, X, y):
             y_new.index = new_index
 
             # Add the new samples to the old dataframe
-            X, y = X.append(X_new), y.append(y_new)
+            # X, y = X.append(X_new), y.append(y_new)
```
tvdboom, line 588: don't comment it out. just replace the old line with the two new ones. Also, the comment isn't necessary. You don't need to comment on changes in syntax.

requirements.txt
```diff
@@ -3,7 +3,8 @@ ipython>=5.5.0
 ipywidgets>=7.6.5  # required by pycaret.internal.display
 tqdm>=4.62.0  # required by pycaret.internal.display
 numpy>=1.21, <1.27
-pandas>=1.3.0, <2.0.0  # Can't >=2.0.0 because of sktime
+pandas>=1.3.0, <2.1.0
+# pandas>=2.1.0; python_version >= "3.9"
```
tvdboom, line 7: why this line? we no longer support Python 3.8 so this is obsolete.

requirements.txt
```diff
@@ -3,7 +3,8 @@ ipython>=5.5.0
 ipywidgets>=7.6.5  # required by pycaret.internal.display
 tqdm>=4.62.0  # required by pycaret.internal.display
 numpy>=1.21, <1.27
-pandas>=1.3.0, <2.0.0  # Can't >=2.0.0 because of sktime
+pandas>=1.3.0, <2.1.0
```
tvdboom, line 6: Why <2.1.0, can't we  make it support the latest version as well? If the problem is the matplotlib version, try checking what the issue is there and bump matplotlib as well.

tvdboom (changes requested):

requirements.txt
celestinoxp: I think python 3.7 is obsolete, but we support python 3.8.
Pandas 2.1 only supports at minimum python 3.9, then who have python 3.8 cannot use pandas 2.1 only 2.0

requirements.txt
celestinoxp: initially I thought about supporting pandas 2.1, but it seems that many problems can arise with third-party packages and in testing, so I thought I would do it in a phased manner, a pull-request for pandas 2.0 and once approved, make another pull-request for pandas 2.1.

requirements.txt
tvdboom: you are right, we do support 3.8, I was confused with another project. this line can be removed then instead of commented.

tvdboom (approved):

PR #3838, opened 2023-12-04
Fix create_docker()

rafinhaLQ:
# Related Issue or bug

When running the generated API on docker would throw this exception:
`pydantic.errors.PydanticUserError: A non-annotated attribute was detected`

And also, I wasn't able to access the API running on docker.

Did not found any open issue on this problems, I found the problems when I was using pycaret.

I changed only `pycaret\internal\pycaret_experiment\tabular_experiment.py`.

There was 4 modifications:
* I added an empty line at the end of the generated Dockerfile and requirements.txt, for good practices;
* Added `pydantic==1.10.0` requirement at generated requirement.txt.
It was necessary because airflow version it's compatible with pydantic version 1.10. If you don't specify the version it will use pydantic version < 2.0.
You could also solve that error without specifying pydantic version changing the requirement `pycaret` to `pycaret[full]`. But it would make your docker image size significantly bigger. That's why I choose to follow this approach. In my tests I made version 1 with `pycaret[full]` and version 2 with `pydantic==1.10.0`. Here's the result:
image
* I also added `apt-get clean` at `RUN` on the generated docker file.
This will remove the cache after install packages, not storing an index, reducing the size of your docker image and speeding up your deployments.
* And finally. I changed the `CMD` on generated dockerfile to `["uvicorn", "{API_NAME}:app", "--host", "0.0.0.0", "--port", "{PORT}"]`.
When running fast api on a container you need to tell Uvicorn to not care about the incoming host IP with option `--host 0.0.0.0`. Without it you will not be able to access your API outside Docker.

# Type of change

# How Has This Been Tested?

I built docker images using python:3-8-slim to python:3.11-slim, in each version of python I built images with and without specifying pydantic version, and with and without telling Uvicorn to not care about incoming host IP.

NA

# Checklist:

# Screenshots

 Original           | Updated
 :--------------------: | :--------------------:
image | image

### DIFF
--- a/pycaret/internal/pycaret_experiment/tabular_experiment.py
+++ b/pycaret/internal/pycaret_experiment/tabular_experiment.py
@@ -2679,7 +2679,6 @@ RUN pip install -r requirements.txt
 
 EXPOSE {PORT}
 
-CMD ["python", "{API_NAME}.py"]
 CMD ["uvicorn", "{API_NAME}:app", "--host", "0.0.0.0", "--port", "{PORT}"]
 
 """.format(
--- a/pycaret/internal/pycaret_experiment/tabular_experiment.py
+++ b/pycaret/internal/pycaret_experiment/tabular_experiment.py
@@ -2654,7 +2654,7 @@ if __name__ == "__main__":
 pycaret
 fastapi
 uvicorn
-pydantic==1.10.0
+pydantic<2.0.0. # required for airflow
 
 """
         print("Writing requirements.txt")
--- a/pycaret/internal/pycaret_experiment/tabular_experiment.py
+++ b/pycaret/internal/pycaret_experiment/tabular_experiment.py
@@ -2654,6 +2654,8 @@ if __name__ == "__main__":
 pycaret
 fastapi
 uvicorn
+pydantic==1.10.0
+
 """
         print("Writing requirements.txt")
         f = open("requirements.txt", "w")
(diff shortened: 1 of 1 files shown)


celestinoxp:
Hi, why pydantic at version 1 instead of using the latest version?

rafinhaLQ:
@celestinoxp If you use pydantic latest version ( < 2.0 ), it will result on a PydanticUserError, cause airflow version used is only compatible if you use a version >= 1.10.0 and < 2.0.0.
You could also use pycaret[full] instead of specifying the pydantic version cause pycaret full will already bring a compatible pydantic version. But I choose to specify the version to make the docker image smaller and faster to build. I put a photo with this two solutions on the pull request description to show it.

Yard1:
Thanks for the PR! Can we set pydantic to < 2.0.0 to loosen the restriction a little?

rafinhaLQ:
@Yard1 Thank you for the feedback! I think it makes much more sense just setting pydantic to < 2.0.0, I made some tests in my machine and it seems to work fine

pycaret/internal/pycaret_experiment/tabular_experiment.py
```diff
@@ -2654,6 +2654,8 @@ def create_docker(
 pycaret
 fastapi
 uvicorn
+pydantic==1.10.0
```
Yard1, line 2657:
```suggestion
pydantic<2.0.0. # required for airflow
```

Yard1 (approved):

PR #3902, opened 2024-02-01
Custom path for logging (logs.log)

### LINKED ISSUE #1057: Make logging configurable
Hi,

first of all many thanks for putting together this package! Ever since reading the book by Kuhn and Johnson and working with caret in R I've been wanting an equivalent Python library.

Referencing #315 I wanted to propose to change the way the default logger is configured in [`pycaret.internal.logging`](https://github.com/pycaret/pycaret/blob/master/pycaret/internal/logging.py#L26):
```python
    try:
        ch = logging.FileHandler("logs.log")    # <-- don't specify a handler with a fixed filename, here
    except:
        print("Could not attach a FileHandler to the logger! No logs will be saved.")
        traceback.print_exc()
        ch = logging.NullHandler()
    ch.setLevel(logging.DEBUG)
```

In [#315](https://github.com/pycaret/pycaret/issues/315#issuecomment-667677236):
> pycaret commented on Aug 2, 2020
>@eonsmile Nope. Is there is a reason you want to disable? Maybe we can mark this as enhancement and somebody can work on making this functionality optional, if that make sense.

For me the use case is an application wrapping a PyCaret model. I've packaged the application using PyInstaller and it is distributed with a setup wizard as a "conventional" Windows application. After installation it's placed in the system's program folder (C:\Program Files (x86)) where normal users don't have write permissions -- but PyCaret wants to create its log file there. As PyCaret is imported in different modules, monkey-patching would require editing some external libraries as well, so re-configuration of the logger appears to be impractical.

Therefore I suggest to make the logging handler configurable, or do you have strong reasons to hard-code the file output?

Additionally, the [Python docs on logging](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) state:
>Note
>
>It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements. 

If you accept this I would be happy to submit a PR implementing it.

Keep up the good work
Joschka

rhsouza:
# Related Issue or bug

The logs.log file is automatically generated in the import script directory (by pycaret/internal/logging.py). This creates problems when the directory is not writable, like AWS Lambda.
I particularly have problems using it directly in the lambda with a container, as the only free directory to write to is '/tmp'.
This update is correctly saving to the desired path.
I found other reports like issue #1057.

Closes #1057

I created a new variable 'logPath' and assigned it an environment variable (PYCARET_CUSTOM_LOGGING_PATH), similar to the one already used for the log level (PYCARET_CUSTOM_LOGGING_LEVEL).
I changed the name of the file ‘logs.log’ to this variable, which takes the path informed in the environment variable and if it doesn’t exist, it continues as ‘logs.log’

# Type of change

# How Has This Been Tested?

After updating the function, I ran pycaret update and provided a new path to the 'PYCARET_CUSTOM_LOGGING_PATH' environment variable.
I imported the functions 'from pycaret.regression import load_model, predict_model' and verified the log file creation as desired.

`NA`

# Checklist:

### DIFF
--- a/pycaret/internal/logging.py
+++ b/pycaret/internal/logging.py
@@ -138,7 +138,7 @@ def create_logger(
     if logger.hasHandlers():
         logger.handlers.clear()
 
-    # custom path to logs.log
+    # custom path for logging, if not given, will use local logs.log
     logPath = os.getenv("PYCARET_CUSTOM_LOGGING_PATH", "logs.log")
     path = logPath if isinstance(log, bool) else log
     ch: Optional[Union[logging.FileHandler, logging.NullHandler]] = None
--- a/pycaret/internal/logging.py
+++ b/pycaret/internal/logging.py
@@ -138,7 +138,9 @@ def create_logger(
     if logger.hasHandlers():
         logger.handlers.clear()
 
-    path = "logs.log" if isinstance(log, bool) else log
+    # custom path to logs.log
+    logPath = os.getenv("PYCARET_CUSTOM_LOGGING_PATH", "logs.log")
+    path = logPath if isinstance(log, bool) else log
     ch: Optional[Union[logging.FileHandler, logging.NullHandler]] = None
     try:
         ch = logging.FileHandler(path)


Yard1 (approved):

Yard1:
Thanks!

PR #3921, opened 2024-02-26
support polars

celestinoxp:
This pull request introduces an initial solution for handling Polars integration into PyCaret code. Initially, completely replacing the code to use Polars natively may seem like a complex and laborious task. As an intermediate approach, this pull request proposes adding conversion capabilities from DataFrame Polars to DataFrame Pandas at certain points in the code.

Converting a dataset from Polars to Pandas is a workaround that allows PyCaret to work with datasets provided in Polars. This approach offers a viable way to continue Polars integration development and exploration without disrupting your existing workflow.

We welcome any feedback or suggestions to improve this solution and are open to discussing more efficient and robust alternatives for integrating Polars with PyCaret.

Related Issue: #3919


Yard1:
Wouldn't converting to pandas remove any benefit from using polars in the first place?

celestinoxp:
> Wouldn't converting to pandas remove any benefit from using polars in the first place?

i can work on this, but i'm not sure how can we do... any ideas?

celestinoxp:
@keitazoumana Can you help us know how we can use dataset type checking without having to repeat thousands of lines of code? Thanks

moezali1:

PR #3923, opened 2024-02-27
Upgrade old evidently api to new api

celestinoxp:
I recently started seeing a problem in pycaret tests, ModuleNotFoundError: No module named 'evidently.dashboard'
After investigating, I discovered that the code used in Pycaret to use evidetly is very old, so it was necessary to update the Pycaret code.
For better context: I was unable to use an old version of evidently because it forced me to change all pycaret dependencies to older versions, in other words, a conflict of dependencies was generated. The only solution is to use recent (and modern) versions

I also asked for help at https://github.com/evidentlyai/evidently/issues/1010

### DIFF
--- a/pycaret/internal/pycaret_experiment/supervised_experiment.py
+++ b/pycaret/internal/pycaret_experiment/supervised_experiment.py
@@ -5658,6 +5658,7 @@ class _SupervisedExperiment(_TabularExperiment):
         return filename
         """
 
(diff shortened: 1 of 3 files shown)


celestinoxp:
@Yard1 can you review this pull-request?

Yard1:
Can we make sure to pin evidently to a working version with those changes?

Issue #3960, opened 2024-04-09
[BUG]: TypeError: descriptor '__call__' for 'type' objects doesn't apply to a 'property' object


amotl:
### pycaret Version

3.3.0

### Python Version

3.11.9

### Description

After downgrading to Joblib 1.3

- GH-3959

when running that code
```python
from pycaret.time_series import (
    setup,
    compare_models,
    tune_model,
    blend_models,
    finalize_model,
    save_model,
)
```

We are receiving that stacktrace.

```python
==================================== ERRORS ====================================
________ ERROR collecting automl_timeseries_forecasting_with_pycaret.py ________
automl_timeseries_forecasting_with_pycaret.py:8: in <module>
    from pycaret.time_series import (
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pycaret/time_series/__init__.py:1: in <module>
    from pycaret.time_series.forecasting.functional import (
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pycaret/time_series/forecasting/functional.py:22: in <module>
    from pycaret.time_series.forecasting.oop import TSForecastingExperiment
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/pycaret/time_series/forecasting/oop.py:18: in <module>
    from sktime.forecasting.base import BaseForecaster, ForecastingHorizon
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/sktime/__init__.py:7: in <module>
    from sktime.utils._maint._show_versions import show_versions
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/sktime/utils/__init__.py:4: in <module>
    from sktime.utils.plotting import plot_series, plot_windows
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/sktime/utils/plotting.py:14: in <module>
    from sktime.datatypes import convert_to
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/sktime/datatypes/__init__.py:5: in <module>
    from sktime.datatypes._check import (
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/sktime/datatypes/_check.py:32: in <module>
    from sktime.datatypes._hierarchical import check_dict_Hierarchical
[...]

PR #3963, opened 2024-04-13
Fix dependency: Python 3.11.9 breaks Dask

amotl:
## Problem
The recent release of Python 3.11.9 introduced an incompatibility with current versions of Dask.

Previous versions of Dask, which is pulled in transitively via Fugue through `pycaret[parallel]`, stopped working on Python 3.11.9, so that needs an update.

## Solution
In order to fix the problem without much efforts, this patch updates to a recent version of Dask>=2024.4.1, which includes a corresponding fix.

## Evaluation

## References

- https://github.com/pycaret/pycaret/issues/3960
- https://github.com/dask/dask/issues/11038

### DIFF
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -36,6 +36,8 @@ m2cgen>=0.9.0  # For model conversion
 evidently~=0.4.16
 
 # Parallel
+dask>=2024.4.1            # Python 3.11.9 breaks Dask (https://github.com/dask/dask/issues/11038)
+distributed>=2024.4.1     # Python 3.11.9 breaks Dask (https://github.com/dask/dask/issues/11038)
 fugue
 flask
 Werkzeug>=2.2,<3.0


requirements-optional.txt
```diff
@@ -36,6 +36,8 @@ m2cgen>=0.9.0  # For model conversion
 evidently~=0.4.16
 
 # Parallel
+dask>=2024.4.1            # Python 3.11.9 breaks Dask (https://github.com/dask/dask/issues/11038)
+distributed>=2024.4.1     # Python 3.11.9 breaks Dask (https://github.com/dask/dask/issues/11038)
 fugue
 flask
 Werkzeug>=2.2,<3.0
```
amotl, line 43: > Previous versions of Dask, which is pulled in transitively via Fugue through `pycaret[parallel]`, stopped working on Python 3.11.9, so that needs an update.

Of course, this should actually go into Fugue, but can also be fixed on both projects, depending on how fast this would converge into corresponding bugfix releases on the one project, or the other, and how convenient package maintainers would like to have the installation process for their users.

So, in the same spirit, that patch submits the same updates to the Fugue project.

- https://github.com/fugue-project/fugue/pull/537

Yard1:
Thanks for the PR, this will be included in https://github.com/pycaret/pycaret/pull/3968

Issue #3959, opened 2024-04-08
[BUG]: Can't use RegressionExperiment with the latest version of joblib (1.4.0)


rfsan:
### pycaret version checks

- [X] I have checked that this issue has not already been reported [here](https://github.com/pycaret/pycaret/issues).

- [X] I have confirmed this bug exists on the [latest version](https://github.com/pycaret/pycaret/releases) of pycaret.

- [X] I have confirmed this bug exists on the master branch of pycaret (pip install -U git+https://github.com/pycaret/pycaret.git@master).

### Issue Description

Pycaret uses the private function `_format_load_msg` from joblib. This function was removed from joblib version 1.4.0

### Reproducible Example

```python
from pycaret.regression import RegressionExperiment
```

### Expected Behavior

Import works correctly

### Actual Results

```python-traceback
22 from typing import TYPE_CHECKING, Union
     24 from joblib.hashing import Hasher, Pickler
---> 25 from joblib.memory import (
     26     MemorizedFunc,
     27     MemorizedResult,
     28     Memory,
     29     NotMemorizedResult,
     30     _format_load_msg,
     31     filter_args,
     32     format_call,
     33     format_signature,
     34     format_time,
     35     get_func_name,
     36 )
     37 from xxhash import xxh128 as xxh
     39 try:

ImportError: cannot import name '_format_load_msg' from 'joblib.memory'
```

### Installed Versions

<details>
System:
    python: 3.11.8 (main, Feb 28 2024, 07:10:00) [GCC 11.4.0]
   machine: Linux-6.8.0-76060800daily20240311-generic-x86_64-with-glibc2.35

PyCaret required dependencies:
                 pip: 24.0
          setuptools: 69.2.0
             pycaret: 3.3.0
             IPython: 8.23.0
          ipywidgets: 8.1.2
                tqdm: 4.66.2
               numpy: 1.26.4
              pandas: 2.1.4
              jinja2: 3.1.3
               scipy: 1.11.4
              joblib: 1.4.0
             sklearn: 1.4.1.post1
                pyod: 1.1.3
            imblearn: 0.12.2
   category_encoders: 2.6.3
            lightgbm: 4.3.0
               numba: 0.59.1
            requests: 2.31.0
          matplotlib: 3.7.5
          scikitplot: 0.3.7
         yellowbrick: 1.5
              plotly: 5.20.0
    plotly-resampler: Not installed
             kaleido: 0.2.1
           schemdraw: 0.15
[...]

PR #3965, opened 2024-04-13
Fix dependency: Joblib 1.4.0 breaks PyCaret

amotl:
## Problem
PyCaret uses the private function `joblib.memory._format_load_msg` from Joblib, which is no longer present in Joblib 1.4.
- https://github.com/pycaret/pycaret/issues/3959

## Solution
In order to fix the problem without much efforts, this patch vendorizes the corresponding function from Joblib without much ado.

## Evaluation

## Install
In order to use the patch instantly before upstream maintainers ran a new bugfix release, if it's important to you, you may want to use this pip package specification to install the package in an ad hoc manner, including the fixes for both dependency flaws.

```console
pip install --upgrade 'pycaret[parallel] @ git+https://github.com/crate-workbench/pycaret@fix-joblib-vendorize'
```

It also works within package metadata specifications like setup.py, setup.cfg, pyproject.toml, or requirements.txt files, with recent versions of pip, when the git command is available on the installation environment. This satisfies most workstation installations, the Google Colab runtime environment, and probably a few others where git is provided.

## Trivia

Please note this patch is not sufficient to make PyCaret work on Python 3.11.9, which also has been released recently [^1], and introduces another incompatibility. You will additionally need that patch to make it work:

- GH-3963

However, this does not impact users who did not update their Python 3.11 version yet, and others who might still be on Python 3.10 or earlier.

[^1]: That's the reason why CI still fails, see GH-3966, which summarizes both issues.

### DIFF
--- a/pycaret/internal/memory.py
+++ b/pycaret/internal/memory.py
@@ -27,7 +27,6 @@ from joblib.memory import (
     MemorizedResult,
     Memory,
     NotMemorizedResult,
-    _format_load_msg,
     filter_args,
     format_call,
     format_signature,
@@ -35,6 +34,8 @@ from joblib.memory import (
     get_func_name,
 )
 from xxhash import xxh128 as xxh
+
+from pycaret.utils.compat import joblib_format_load_msg as _format_load_msg
 
 try:
     from math import prod
--- a/pycaret/utils/compat.py
+++ b/pycaret/utils/compat.py
@@ -1,0 +1,31 @@
+import os
+import time
+
+from joblib.logger import format_time
+
+
+def joblib_format_load_msg(func_id, args_id, timestamp=None, metadata=None):
+    """
+    Helper function to format the message when loading the results.
+    Vendorized from Joblib 1.3.2 to retain compatibility with Joblib 1.4.0.
+    """
+    signature = ""
+    try:
+        if metadata is not None:
+            args = ", ".join(
+                [
+                    "%s=%s" % (name, value)
+                    for name, value in metadata["input_args"].items()
+                ]
+            )
+            signature = "%s(%s)" % (os.path.basename(func_id), args)
+        else:
+            signature = os.path.basename(func_id)
+    except KeyError:
+        pass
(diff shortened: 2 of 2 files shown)


celestinoxp:
@amotl tests are failing... you must fix the remaining errors...

amotl:
> @amotl tests are failing... you must fix the remaining errors...

Thanks for noticing. We did not see that error before, apologies. We will look into it.

```python
AttributeError: 'FastMemorizedFunc' object has no attribute '_get_output_identifiers'
```
-- https://github.com/pycaret/pycaret/actions/runs/8676442808/job/23790959421?pr=3965#step:6:1611

celestinoxp:
ping @Yard1

amotl:
It looks like the error around `FastMemorizedFunc` might be yet another thing, only happening on Python 3.9 and 3.10. The error on Python 3.11 will be remedied by GH-3963. We will look into where the new flaw may be coming from.

- Python 3.9: `AttributeError: 'FastMemorizedFunc' object has no attribute '_get_output_identifiers'`
  https://github.com/pycaret/pycaret/actions/runs/8676442808/job/23790959421?pr=3965#step:6:2968

- Python 3.10: `AttributeError: 'FastMemorizedFunc' object has no attribute '_get_output_identifiers'`
  https://github.com/pycaret/pycaret/actions/runs/8676442808/job/23790959545?pr=3965#step:6:2968

- Python 3.11: `TypeError: descriptor '__call__' for 'type' objects doesn't apply to a 'property' object`
  https://github.com/pycaret/pycaret/actions/runs/8676442808/job/23790959722?pr=3965#step:6:78

amotl:
Hi again,

it looks like the errors with Python 3.9 and 3.10 are also related to Joblib 1.4.0, so we are abstaining from vendorizing, and favor downgrading as a quick measure.

In this spirit, we diverted this into another patch...
- GH-3967

... and will close this one again.

With kind regards,
Andreas.

Issue #3966, opened 2024-04-13
[META]: Joblib 1.4.0 breaks PyCaret. Python 3.11.9 breaks Dask.


amotl:
## About
This issue bundles and describes two flaws with dependencies, which have been occurring at the same time, five days ago, but are independent from each other.

## What's Inside
Enumerating the incompatibilities with recent releases in Python.

### Joblib 1.4.0 incompatibility
PyCaret is incompatible with Joblib 1.4.0, because it uses a private function which is no longer there.

- https://github.com/pycaret/pycaret/issues/3959
- ~~https://github.com/pycaret/pycaret/pull/3965~~
- https://github.com/pycaret/pycaret/pull/3967

### Python 3.11.9 incompatibility
Previous versions of Dask, which is pulled in transitively via Fugue through `pycaret[parallel]`, stopped working on Python 3.11.9, so that needs an update.

- https://github.com/dask/dask/issues/11038
- https://github.com/pycaret/pycaret/issues/3960
- https://github.com/pycaret/pycaret/pull/3963

## Evaluation
It is advisable to release this update on behalf of a bugfix release, because otherwise, on new and fresh installations, pulling in Joblib 1.4.0, or when upgrading to Python 3.11.9 and newer, things will go south.

### Timeline
Both flaws became visible on downstream CI/GHA jobs effectively at the same time, caused by those events.

- Joblib 1.4.0 was released on Apr 8, 2024, 15:08 GMT.
  https://github.com/joblib/joblib/releases/tag/1.4.0

- GitHub updated their runners to use Python 3.11.9 on Apr 8, 2024, 15:31 GMT+2.
  https://github.com/actions/runner-images/pull/9656

### pycaret Version

3.3.0

### Python Version

Python 3.11

amotl:
PyCaret 3.3.1 has been released, most probably unbreaking those details mentioned above. Thank you very much.

-- https://github.com/pycaret/pycaret/releases/tag/3.3.1

amotl:
Verified successfully by upgrading to PyCaret 3.3.1.

- https://github.com/crate/cratedb-examples/pull/421

Issue #3964, opened 2024-04-13
[TESTS]: `IndexError: index 18 is out of bounds for axis 0 with size 18`


amotl:
### Problem

When running the software tests on GH-3963, it looks like there is an unrelated error on the test suite outcome, coming from plotting results of the `pycaret.classification` subsystem, raised by `test_plot`.

```
FAILED tests/test_classification_plots.py::test_plot - IndexError: index 18 is out of bounds for axis 0 with size 18
```

- https://github.com/pycaret/pycaret/actions/runs/8675819351/job/23789656626?pr=3963#step:6:120
- https://github.com/pycaret/pycaret/actions/runs/8675819351/job/23789656871?pr=3963#step:6:120
- https://github.com/pycaret/pycaret/actions/runs/8680898663/job/23802476303?pr=3967#step:6:120

amotl:
This error apparently currently blocks GH-3967 on CI.

PR #3967, opened 2024-04-14
Fix dependency: Joblib 1.4.0 breaks PyCaret, so downgrade to <1.4

amotl:
## Problem
PyCaret uses the private functions `joblib.memory._format_load_msg` and `FastMemorizedFunc._get_output_identifiers` from Joblib, which are no longer present in Joblib 1.4.

- https://github.com/pycaret/pycaret/issues/3959

## Solution
In order to fix the problem without much efforts, this patch just downgrades to Joblib 1.3.x without further ado.

## Evaluation

## Install
In order to use the patch instantly before upstream maintainers ran a new bugfix release, if it's important to you, you may want to use this pip package specification to install the package in an ad hoc manner, including the fixes for both dependency flaws.

```console
pip install --upgrade 'pycaret[parallel] @ git+https://github.com/crate-workbench/pycaret@fix-joblib-downgrade'
```

It also works within package metadata specifications like `setup.py`, `setup.cfg`, `pyproject.toml`, or `requirements.txt` files, with recent versions of pip, when the git command is available on the installation environment. This satisfies most workstation installations, the Google Colab runtime environment, and probably a few others where git is provided.

## Trivia

Please note this patch is not sufficient to make PyCaret work on Python 3.11.9, which also has been released recently [^1], and introduces another incompatibility. You will additionally need that patch to make it work:

- GH-3963

However, this does not impact users who did not update their Python 3.11 version yet, and others who might still be on Python 3.10 or earlier.

[^1]: That's the reason why CI still fails, see GH-3966, which summarizes both issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,7 @@ trained_models/
 mlflow_backend.py
 demo4.py
 dask*/
-/.venv
+/.venv*
 /.devcontainer
 tmp/
 nbtest1.ipynb
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ numpy>=1.21, <1.27
 pandas<2.2.0 #test_check_fairness_regression fails with pandas 2.2.0
 jinja2>=3
 scipy>=1.6.1,<=1.11.4 #to fix later (https://github.com/mljar/mljar-supervised/issues/691)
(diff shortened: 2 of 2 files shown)


amotl:
CI raises `IndexError: index 18 is out of bounds for axis 0 with size 18` here. It might be unrelated to our change, so we also reported it seperately.

- GH-3964

Can you make any sense of it?

Yard1:
Thanks, merged this and fixing CI separately.

PR #4033, opened 2024-07-24
refactor code with flake8-bugbear

celestinoxp:
update code
fix small bugs


celestinoxp:
@AyushExel  tried updating pycaret/loggers/base_logger.py with a suggestion from flake8-bugbear, when adding "abstractmethod" errors appeared in mlflow... could you help?

moezali1:

PR #4041, opened 2024-07-30
Reset index after concatenating train and test dataframes

### LINKED ISSUE #4039: [ENH]: Ease index requirement for provided test_data

Currently if we provide a Pandas DataFrame to `test_data` in the `setup()` function and its indices have overlap with those in `self.data` of the `Preprocessor` instance (mostly like by setting the `data` keyword), the concatenated dataframe as below can have overlapping indices. 
https://github.com/pycaret/pycaret/blob/1ef3451ec9296e2b770c7c86e97533095bf2602b/pycaret/internal/preprocess/preprocessor.py#L261 
It will trigger the error from https://github.com/pycaret/pycaret/blob/1ef3451ec9296e2b770c7c86e97533095bf2602b/pycaret/internal/preprocess/preprocessor.py#L207 in the `set._set_index` function.

It's quite common that a test dataframe is generated INDEPENDENTLY along with the training data, so it's better to easy the index requirement for `test_data`. 

### Describe your proposed solution

Just change `self.data = self._set_index(pd.concat([self.data, test_data]))` to `self.data = self._set_index(pd.concat([self.data, test_data]).reset_index(drop=True))`

We can apply `self._set_index` to `test_data` first by setting the indices of `test_data` according to `set.data.index`. It requires change in that function. This solution could have better compatibility.

### Additional context

I am not sure why that index duplicate checking line was added but if the purpose was to make sure the `test_data` has no overlap with the train data, we can check the data duplication before concatenation and raise an error/warning if there are any.

Ruibin-Liu:
# Related Issue or bug

Info about Issue or bug

Closes #4039

Simply added `.reset_index(drop=True)` after the `pd.concat` operation.  

# Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

NA 

NA

# Checklist:

# Screenshots

 Original           | Updated

### DIFF
--- a/pycaret/internal/preprocess/preprocessor.py
+++ b/pycaret/internal/preprocess/preprocessor.py
@@ -258,7 +258,7 @@ class Preprocessor:
                 self.data.index = self.index[: len(self.data)]
                 test_data.index = self.index[-len(test_data) :]
 
-            self.data = self._set_index(pd.concat([self.data, test_data]))
+            self.data = self._set_index(pd.concat([self.data, test_data]).reset_index(drop=True))
             self.idx = [
                 self.data.index[: -len(test_data)],
                 self.data.index[-len(test_data) :],


Yard1 (approved):
Thanks!

PR #4124, opened 2025-02-09
[MNT] migrate packaging to `pyproject`

fkiraly:
This PR migrates packaging from `setup.py` and `requirements` files to `pyproject`.


pyproject.toml
```diff
@@ -0,0 +1,236 @@
...
+[project.optional-dependencies]
+# there are the following dependency sets:
+# - full - all soft dependencies
+# - analysis
+# - models
+# - tuners
+# - mlops
+# - parallel
+# - dev
+# - test
```
amotl, line 98: Thank you for staging this. You may also want to absorb the gist of that patch into this one?

- GH-4126

fkiraly: This PR only does a pure refactor.

For changing the dep sets, I think we need to isolate the intended soft deps in the code first, or the package may break!

See more details here: https://github.com/pycaret/pycaret/pull/4126#pullrequestreview-2604481275

amotl: GH-4129 is much better. Thanks!

fkiraly: I will give you maintainer access to the fork, so you can do both.

ngupta23 (approved):

Issue #4121, opened 2025-02-08
[ENH]: Support for Python 3.13


amotl:
### Describe the feature you want to add to this project

Hi. Coming from GH-4054, let me enumerate blockers for updating to Python 3.13 here, as discovered per CI runs on https://github.com/crate/academy-fundamentals-course/pull/23.

- https://github.com/numba/llvmlite/issues/1084
- https://github.com/alkaline-ml/pmdarima/issues/588

### Describe your proposed solution

Pick up maintenance again.

- https://github.com/pycaret/pycaret/issues/4106
- https://github.com/pycaret/pycaret/issues/4119

### Describe alternatives you've considered, if relevant

Migrate to AutoGluon, as proposed by others.

fkiraly:
I would propose to address this by clean soft dependency management and isolation.

`pmdarima` looks like it is used only in the time series module, so we should isolate it as a soft dependency.

More generally, a review of the soft dependency strategy would be a priority I would suggest, the package has too many core dependencies and the soft dependencies are not cleanly isolated.

amotl:
With "isolating soft dependencies", you are referring to the process of breaking out certain dependencies into "extra" dependencies, in order to make them optional for the fundamental package installation?

fkiraly:
> With "isolating soft dependencies", you are referring to the process of breaking out certain dependencies into "extra" dependencies, in order to make them optional for the fundamental package installation?

@amotl, yes - I would go even further, having multiple, more granular sets of soft dependencies. So someone who primarily wants to use the time series module with extras does not need to install the extras for classification etc.

amotl:
Hi. What do you think about this proposal to isolate dependencies, in order to be able permit installing and using PyCaret on most recent Python 3.13, @moezali1, @Yard1, @tvdboom, and @ngupta23?

amotl:
This patch aims to unblock installing PyCaret on Python 3.13.

- GH-4126

fkiraly:
In-principle sensible, but I think we should not skip the usual sequence in isolating dependencies:

1. isolate dependencies in the code, i.e., ensure the package does not break without them installed - only the feature
2. optionally, give users an advance warning that this will change
3. move the dependency to an optional dep set

Further, `pmdarima` - the only one blocking 3.13 - is imported only in two small locations, while the other time series dependencies are entagled more significantly with the code base.

I would hence suggest to start with that: https://github.com/pycaret/pycaret/issues/4127

[...]

PR #4126, opened 2025-02-09
Dependencies: Isolate "timeseries" dependencies into separate "extra"

amotl:
Hi there,

this patch implements the idea to also isolate dependencies concerning "timeseries" functionality. It also defines that the pmdarima package will only be installed up until Python 3.12, because it is not available for Python 3.13, yet.

The intent is to unblock installing PyCaret on Python 3.13. Thank you in advance for looking into this. ✨ 

With kind regards,
Andreas.

## References
- https://github.com/pycaret/pycaret/issues/4121
- https://github.com/alkaline-ml/pmdarima/issues/588

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ pip install pycaret[mlops]
 # install parallel extras
 pip install pycaret[parallel]
 
+# install timeseries extras
+pip install pycaret[timeseries]
+
 # install test extras
 pip install pycaret[test]
 
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -38,3 +38,9 @@ distributed>=2024.4.1     # Python 3.11.9 breaks Dask (https://github.com/dask/d
 fugue~=0.8.0
 flask
 Werkzeug>=2.2,<3.0
+
+# Time-series
+statsmodels>=0.12.1
+sktime #sktime adds support for scikit-learn 1.5 in version sktime 0.31
+tbats>=1.1.3
+pmdarima>=2.0.4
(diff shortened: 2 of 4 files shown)


fkiraly (changes requested):
This is a sensible change, but I think users should be given prior warning.

Imagine a user with a deployment pipeline that just installs the base deps, this would break after this change is released.

Compare the `sktime` deprecation policy and the rationale of "working code should not break without clear warning":
https://www.sktime.net/en/stable/developer_guide/deprecation.html

There is also an even more serious potential "worst case" where the dependencies are not sufficiently isolated in the code base, and the package might break if installed only the basic dep set.

(hypothetically: somewhere you have an `import tbats` or similar, at module level, and this is triggered by loading `pycaret`.)

So, may I suggest to do this *only* after we actually go through the code base and isolate the dependencies?

I would suggest to do that next, anyway - trying to isolate core dependencies, and then start raising warnings in all cases where features are used from imports that are scheduled to move out of the core dep set.

amotl:
Thanks for providing a solid plan for isolating this group of dependencies. I admit it was just a quick shot to probe it on CI, and needs more love along the lines of what you are suggesting.

fkiraly:
Suggestion: you could look at `pmdarima` - this has only two imports that I can find. `sktime` is all over the place, on module level and also in places that are not time series. `statsmodels` is somewhere in the middle.

Perhaps it makes sense to focus on `pmdarima` first, and try to isolate - or even replace - the two imports?

fkiraly:
Opened issue here: https://github.com/pycaret/pycaret/issues/4127

fkiraly:
Example on how to isolate dependencies inside the code base: https://github.com/pycaret/pycaret/pull/4129

amotl:
Closing this in favor of GH-4129. Thank you very much, @fkiraly.

PR #4141, opened 2025-02-15
[ENH] deduplicate dependency checkers and isolate soft dependency imports by using `skbase`

fkiraly:
This PR deduplicates dependency checkers by using `skbase`, which is currently an implied dependency, and which contains equivalent functionality.

This improves maintainability of the dependencies module through deduplication.

Importantly, it also helps with soft dependency isolation because the `skbase` dependency checker does not actually import all modules while checking - making future work on reducing impact of soft dependencies easier.

In line with best practice, `packaging` and `scikit-base` are added as dependencies to `pyproject.toml`, since there are now direct imports.

No deprecation or change period is necessary, since both packages were already implied by core dependencies.

Requires https://github.com/pycaret/pycaret/pull/4124 because we are using `pyproject.toml`.


ngupta23:
I have merged #4124. Please fix the conflicts so we can merge this as well. Thanks!

fkiraly:
conflicts fixed - failures afaik are not caused by this but are pre-existing on main

moezali1:

PR #4145, opened 2025-02-17
[MNT] Bump SHAP version to 0.46.0

celestinoxp:
Bump SHAP version to 0.46.0

Closes https://github.com/pycaret/pycaret/issues/4149

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,7 +104,7 @@ dependencies = [
 # users can install via "pip install pycaret[full]"
 #
 full = [
-  "shap>=0.44.0,<=0.46.0",
+  "shap>=0.44.0,<0.47.0",
   "interpret>=0.2.7",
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Can you explain the change?

The pre-PR version says that `shap` must *not* be 0.44.0. Now you allow 0.44.0, and higher versions.

Your PR text says that older versions broke the installation, but I see no evidence for:

* a broken installation in pre-PR CI
* that older versions would have been installed in pre-PR CI

Can you kindly explain?

fkiraly (changes requested):
As discussed in https://github.com/pycaret/pycaret/issues/4149: you are right about the notation, I was in error.

As a change though, may I request an upper bound of <0.47.0?

ngupta23 (approved):

PR #4147, opened 2025-02-17
fix tests

celestinoxp:
many packages are in new versions but the pycaret code has not followed these changes. You need to start over with the old packages and update the pycaret code until it supports the latest versions.

example: scikit-learn had no limitations so as the package is in version 1.6 pycaret does not pass the tests, so it is necessary to downgrade to 1.5, modify the code, for the tests to pass and only then update again to 1.6 and update the pycaret code.

### DIFF
--- a/pycaret/internal/preprocess/preprocessor.py
+++ b/pycaret/internal/preprocess/preprocessor.py
@@ -206,7 +206,7 @@ class Preprocessor:
 
         if df.index.duplicated().any():
             raise ValueError(
-                "Invalid value for the index parameter. There are duplicate indices "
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
Can you explain quickly which test in which CI element is failing?

I think this PR is problematic as long as you are changing tests *and* version bounds. Because if the reason is the new sklearn version (which is plausible), then downgrading should already fix the test?

If you are fixing two tests, I would do that in two separate PR.

Further, could you ensure you have descriptive PR titles? E.g., which tests are you fixing, instead of "fix tests". If you are downgrading a version, that should also have a descriptive Pr title.

PR #4150, opened 2025-02-20
[WIP][MNT] Downgrade package versions to last stable dev point and update gradually

celestinoxp:
PyCaret is not compatible with the latest versions of its dependencies, as development has been stalled for a long time. To address this, a strategy is proposed to resume development in a controlled manner.

**Strategy:**
1. **Pin dependency versions**: Restrict dependencies to a specific point (e.g., ~1 year ago), such as `sktime>=0.31,<0.32`, to prevent newer, incompatible versions (e.g., `0.36`) from being installed and causing errors.
2. **Run tests**: Ensure GitHub CI tests pass with these pinned versions to establish a stable baseline.
3. **Gradual updates**: Incrementally update dependencies, adapting PyCaret's code as needed and verifying test results at each step.

This PR starts by downgrading key dependencies to their last known compatible versions and will evolve as the codebase is brought up to date.


celestinoxp:
@amotl Would you be available to lend a hand here? I'm having trouble understanding what's happening with the tests. joblib does not display an explanatory message that allows you to get to the root of the problem.

amotl:
> **Run tests:** Ensure GitHub CI tests pass with these pinned versions to establish a stable baseline.

Hi @celestinoxp. Thanks a stack for dedicating work to this. 💯 

This PR run [responded like](https://github.com/pycaret/pycaret/actions/runs/13506094493/job/37735864928?pr=4150#step:6:3443):

> This could be caused by a segmentation fault while calling the function or by an excessive memory usage causing the Operating System to kill the worker.

oO. That's probably one of those super hard errors to debug, exaggerated by that it is happening on a remote CI system.

But hey, is it really true that you brought it down to just [1 failed test run](https://github.com/pycaret/pycaret/actions/runs/13506094493/job/37735864928?pr=4150#step:6:3446)? That would be so excellent already!

I don't know what I could contribute to improve the situation here. Recently, I heard colleagues talking about ways to log into GHA's execution context from your workstation, in order to be able to interactively use the environment where the problem happens, at hand. Would you be happy if I would dig out corresponding information, in order to present it here for anyone who would like to have a closer look what's happening within C/GHA/PyCaret at this particular spot where the CI workflow goes south?

Is the test suite actually succeeding on your machine, and just failing on CI/GHA?

amotl:
Hi.

> I will fix everything.

Excellent! 🚀

> It is necessary to mark all packages with the "<" symbol.

I am doing the same with my libraries where I do not aim for unrestricted downstream havocs. 👍 

> The problem is in compatibility with joblib 1.4.

All right. Is it possible to also fix it by limiting the upper version boundary like with the other dependencies?

> If I could get funding, I could dedicate myself full time to PyCaret.

That sounds interesting, and I think we should emphasize it collaboratively. I can also ask about deriving funding from fragments of our next financing round, but I am not a C-level person, so it is not likely I can make a significant difference in convincing others about it.


With kind regards,
Andreas.

[^1]: What I **can** do, is to prioritize any feature requests or bug reports pending at https://github.com/crate/mlflow-cratedb, if you would have a need for them in one way or another. But this does not pay the bills, I know. 🍀

pyproject.toml
```diff
@@ -81,7 +81,7 @@ dependencies = [
   "plotly-resampler>=0.8.3.1",
   # Time-series
   "statsmodels>=0.12.1",
-  "sktime",  # sktime adds support for scikit-learn 1.5 in version sktime 0.31
+  "sktime==0.31.0",
```
fkiraly, line 84: Strongly suggest, do not pin, but only upper bound.

Even a single pin can destroy some user´s setup due to incompatibility.

fkiraly (changes requested):
I would suggest:

* leave lower bounds if they were in the last release
* add upper bounds to last known stable test run
* do not use pins (see above)

More generally, the de-facto policy used in `sktime` is this:
https://github.com/sktime/sktime/issues/1480

I think the defensive policy might also be advisable, due to the larger number of soft deps. But that is a discussion for a later day.

pyproject.toml
celestinoxp: This is temporary, I just fixed it to try to figure out where the test errors are coming from.

amotl:
Hi @celestinoxp. FWIW, I've whipped up your branch into the CI workflow runs of ours, that invoke a few Jupyter Notebooks which use PyCaret.

- Temporary patch: https://github.com/crate/cratedb-examples/pull/844
- Tracking ticket: https://github.com/crate/cratedb-examples/issues/843

<details><summary>More</summary>
<p>

You may want to inspect a failed test run [over here](https://github.com/crate/cratedb-examples/actions/runs/13593537743/job/38005137914?pr=844), errors are mostly revolving around that:
```
AttributeError: 'list' object has no attribute 'get'
```
However, because this happens in `pycaret/loggers/mlflow_logger.py`, and we are using the software together with our [mlflow-cratedb](https://github.com/crate/mlflow-cratedb), which is also not up-to-date, we think it might be a compatibility issue in that regard, so PyCaret should not be concerned about it.
</p>
</details>

amotl:
Hi. Are you able to improve the situation by increasing the recursion limit, using `sys.setrecursionlimit()`?

- https://github.com/joblib/joblib/issues/224

celestinoxp:
> Hi. Are you able to improve the situation by increasing the recursion limit, using `sys.setrecursionlimit()`?
> 
> * [Recursion limit reached during joblib.dump() joblib/joblib#224](https://github.com/joblib/joblib/issues/224)

i tried a temporary solution with setrecursionlimit, but tests/test_time_series_blending.py still not passing...

celestinoxp:
@fkiraly @ngupta23 @amotl @PabloJMoreno I made the necessary changes for the tests to pass. From here on, each change must have its specific PR.
Anyone interested in new corrections or improvements should contact me via slack or email.

amotl:
@celestinoxp: This is so sweet. Thank you for your excellent work on this. 🌻

ngupta23 (approved):

